PR #6, opened 2022-01-14
Equinix Metal provider

DavidSpek:
Adds Equinix Metal as a cloud provider in the CLI.


pkg/provider/constants.go
```diff
@@ -140,3 +141,70 @@ provider "kubernetes" {
...
 }
 `
+
+// TODO: Figure out how to deal with local backend
+// TODO: Figure out how to configure the Kubernetes provider (rke/equinix datasource or store somewhere?)
+const equinixBackendTemplate = `terraform {
+  backend "local" {
+    path = "../../{{ .Values.Bucket }}/{{ .Values.__CLUSTER__ }}/{{ .Values.Prefix }}/terraform.tfstate"
+  }
+{{- if .Values.ClusterCreated }}
```
michaeljguarino, line 151: I feel like it might have been simpler to just configure the helm provider locally in the bootstrapping module.

pkg/provider/equinix.go
```diff
@@ -0,0 +1,278 @@
...
+
+	if err != nil {
+		return utils.ErrorWrap(err, "failed to terminate instance")
+	}
+
+	return nil
+}
+
+func getProjectIDFromName(projectName string) (string, error) {
+	cmd := exec.Command("metal", "project", "get", "-o", "json")
```
michaeljguarino, line 254: Is it possible to get this from the go sdk so we can remove dependency on the metal cli?

There's a backlog task to do that for the other providers, but it's more difficult there since it's somewhat required for fetching access tokens in local kubeconfigs.

DavidSpek: That's probably possible.
The `Decommision` function also doesn't work, since it needs the API token. What is that function used for?

pkg/provider/constants.go
DavidSpek: Probably haha. I was running into some issues with those providers before so that'll need some more attention and testing as well.

pkg/provider/equinix.go
```diff
@@ -0,0 +1,280 @@
...
+	var projectID string
+	var res []struct {
+		Name string
+		Id   string
+	}
+	json.Unmarshal(out, &res)
+
+	for _, project := range res {
+		if project.Name == projectName {
+			projectID = project.Id
```
michaeljguarino, line 272: we should break here rather than continuing the loop

pkg/provider/equinix.go
```diff
@@ -0,0 +1,280 @@
...
+	json, err := runtime.Encode(clientcmdlatest.Codec, config)
+	if err != nil {
+		return err
+	}
+	output, err := yaml.JSONToYAML(json)
+	if err != nil {
+		return err
+	}
+
+	err = ioutil.WriteFile(filepath.Join(usr.HomeDir, ".kube/config"), output, 0644)
```
michaeljguarino, line 193: can just do `return ioutil.WriteFile(...)` here

pkg/provider/equinix.go
```diff
@@ -0,0 +1,280 @@
...
+
+	for _, kubeconfig := range kubeconfigs {
+		mergo.Merge(mapConfig, kubeconfig, mergo.WithOverride)
+	}
+
+	// merge all of the struct values
+	nonMapConfig := clientcmdapi.NewConfig()
+	for i := range kubeconfigs {
+		kubeconfig := kubeconfigs[i]
+		mergo.Merge(nonMapConfig, kubeconfig, mergo.WithOverride)
```
michaeljguarino, line 175: I'm not quite following why this is different to what was done on line 168, should we have not set the mergo.WithOverride opt this time or somethign?

DavidSpek: I think the comment on line 178 and 179 explain why this is necessary. I actually based this off the code from Kubernetes client-go which I was running into an issue with.
https://github.com/kubernetes/client-go/issues/901

michaeljguarino (approved):
ensure workspace.yaml is encrypted, but otherwise good

PR #19, opened 2022-02-23
use provider scaffolds api

ycryptx:
## Summary
This PR utilizes the new Plural gql API's to retrieve available TF providers and their base scaffold code.
Once retrieved, the list of available providers and scaffolds are cached in each execution of CLI command.
Note that now the list providers is capitalized, and the "google" provider is now called "gcp"
closes #167

## Checklist

### DIFF
--- a/pkg/provider/provider.go
+++ b/pkg/provider/provider.go
@@ -2,8 +2,10 @@ package provider
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/pluralsh/plural/pkg/api"
 	"github.com/pluralsh/plural/pkg/config"
 	"github.com/pluralsh/plural/pkg/manifest"
 	"github.com/pluralsh/plural/pkg/utils"
@@ -23,7 +25,15 @@ type Provider interface {
 	Decommision(node *v1.Node) error
 }
 
+type Providers struct {
+	AvailableProviders []string
+	Scaffolds          map[string]string
+}
+
+var providers = Providers{}
+
 func Bootstrap(manifestPath string, force bool) (Provider, error) {
+	GetAvailableProviders()
 	if utils.Exists(manifestPath) {
 		man, err := manifest.Read(manifestPath)
 		if err != nil {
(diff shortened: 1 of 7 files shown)


pkg/provider/provider.go
```diff
@@ -36,8 +46,35 @@ func Bootstrap(manifestPath string, force bool) (Provider, error) {
 	return Select(force)
 }
 
+func GetProviderScaffold(provider string) (string, error) {
```
michaeljguarino, line 49: I think these functions should probably be private not public, but not a huge deal

pkg/provider/provider.go
```diff
@@ -67,22 +104,24 @@ func Select(force bool) (Provider, error) {
...
 	survey.AskOne(prompt, &provider, survey.WithValidator(survey.Required))
 	utils.Success("Using provider %s\n", provider)
 	return New(provider)
 }
 
 func FromManifest(man *manifest.Manifest) (Provider, error) {
-	switch man.Provider {
-	case GCP:
+	switch strings.ToUpper(man.Provider) {
+	case "GCP":
```
michaeljguarino, line 116: I think this is going to have downstream implications, since the current values of these strings are used in configuring applications and other things.  We should certainly not modify them for safety.

ycryptx: Yes it's a bit intrusive of a change

michaeljguarino: What scares me is I'm about 99% sure it'll cause externaldns to be misconfigured, which will bork a lot of cluster setups.  But there's a ton of templating in other apps that are dependent on the existing provider casing.

pkg/provider/provider.go
```diff
@@ -23,7 +25,15 @@ type Provider interface {
...
 
+type Providers struct {
+	AvailableProviders []string
+	Scaffolds          map[string]string
+}
+
+var providers = Providers{}
+
 func Bootstrap(manifestPath string, force bool) (Provider, error) {
+	GetAvailableProviders()
```
michaeljguarino, line 36: considering how this function basically is just called for side-effects, I definitely think it should definitely be made private

michaeljguarino (approved):

PR #30, opened 2022-03-16
GCP allow user to set Region

DavidSpek:
## Summary
This change allows users to configure the GCP region in which they would like to deploy their cluster, bringing it to feature parity with AWS and Azure. It also moves away from using a Zone instead of the Region, since the Zone level setting doesn't seem to be used anywhere and is inconsistent with the other providers on Plural.

This change will also require changes to the CLI template (see ...).

## Checklist

### DIFF
--- a/pkg/provider/constants.go
+++ b/pkg/provider/constants.go
@@ -71,15 +71,9 @@ const gcpBackendTemplate = `terraform {
   }
 }
 
-locals {
-  gcp_location  = {{ .Values.Location | quote }}
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-}
-
 provider "google" {
   project = {{ .Values.Project | quote }}
-  region  = local.gcp_region
+  region  = {{ .Values.Region | quote }}
 }
 
 data "google_client_config" "current" {}
@@ -93,7 +87,7 @@ provider "kubernetes" {
 {{ else }}
 data "google_container_cluster" "cluster" {
   name = {{ .Values.Cluster }}
-  location = local.gcp_region
+  location = {{ .Values.Region | quote }}
 }
 
 provider "kubernetes" {
--- a/pkg/provider/gcp.go
+++ b/pkg/provider/gcp.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"os/exec"
-	"strings"
 
 	v1 "k8s.io/api/core/v1"
 
@@ -24,7 +23,7 @@ type GCPProvider struct {
 	Clust         string `survey:"cluster"`
 	Proj          string `survey:"project"`
 	bucket        string
-	region        string
+	Reg           string `survey:"region"`
 	storageClient *storage.Client
 	ctx           context.Context
 }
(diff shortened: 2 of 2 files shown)


pkg/provider/constants.go
```diff
@@ -71,15 +71,9 @@ const gcpBackendTemplate = `terraform {
...
-locals {
-  gcp_location  = {{ .Values.Location | quote }}
-  gcp_location_parts = split("-", local.gcp_location)
-  gcp_region         = "${local.gcp_location_parts[0]}-${local.gcp_location_parts[1]}"
-}
-
 provider "google" {
   project = {{ .Values.Project | quote }}
-  region  = local.gcp_region
+  region  = {{ .Values.Region | quote }}
```
michaeljguarino, line 76: this should no longer be used I thought?

DavidSpek: These templates are indeed no longer used.

pkg/provider/gcp.go
```diff
@@ -24,7 +23,7 @@ type GCPProvider struct {
 	Clust         string `survey:"cluster"`
 	Proj          string `survey:"project"`
 	bucket        string
-	region        string
+	Reg           string `survey:"region"`
 	storageClient *storage.Client
 	ctx           context.Context
```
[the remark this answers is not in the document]
michaeljguarino: I think i must have been plumbing the context around since it was needed for google clients.  The field itself can likely be dropped and replaced with calls to `context.Background()`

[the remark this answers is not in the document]
DavidSpek: I was wanting to place a bucket region in there actually, since it requires a bit of parsing that will be annoying to do upstream.

pkg/provider/gcp.go
```diff
@@ -109,7 +145,9 @@ func (gcp *GCPProvider) CreateBackend(prefix string, ctx map[string]interface{})
 	}
 
 	ctx["Project"] = gcp.Project()
-	ctx["Location"] = gcp.Region()
+	// Location is here for backwards compatibility
+	ctx["Location"] = getLocation()
```
michaeljguarino, line 149: we probably don't want to shell out to gcloud every time this is run.  Ideally we'd also put this in the context once and fetch it from there.

michaeljguarino: Actually we definitely don't want to be shelling to gcloud here, because it'll break in the console

DavidSpek: Yeah I wasn't sure about that either. I'll put it in the context in the morning, but for testing that change shouldn't make much difference.

pkg/provider/constants.go
michaeljguarino: We should remove then

pkg/provider/gcp.go
```diff
@@ -40,6 +48,11 @@ var gcpSurvey = []*survey.Question{
 		Prompt:   &survey.Input{Message: "Enter the name of its gcp project"},
 		Validate: utils.ValidateAlphaNumeric,
 	},
+	{
+		Name:     "region",
+		Prompt:   &survey.Input{Message: "What region will you deploy to?", Default: "us-east1"},
```
michaeljguarino, line 53: We should make this a Select input, and you can probably just repurpose the list here for supported regions: https://github.com/pluralsh/plural/blob/master/www/src/components/shell/cloud/gcp.js#L9

michaeljguarino: (we'll want to do this for the other clouds too tbh)

michaeljguarino (approved):

PR #31, opened 2022-03-28
Add KinD provider

DavidSpek:
## Summary
Adds support for deploying a local cluster using the KinD provider.


michaeljguarino (approved):
We'll also need to bump the vsn in homebrew-plural

Issue #44, opened 2022-04-08
invalid memory address when building if app is not installed in the API


DavidSpek:
## Summary
If an application is not installed in the Plural API and a user runs the `plural build --only <not-installed-application>`, an invalid memory address error will occur.

## Reproduction
Run `plural build --only ...` for an application that is not installed in the API.

## CLI UI/UX Issue Screenshots
```bash
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x10 pc=0x19b9e57]

goroutine 1 [running]:
main.doBuild(0x1efe9d5, 0x0, 0x6)
        /home/david/Plural/plural-cli/cmd/plural/deploy.go:135 +0x37
main.build(0xc000166dc0)
        /home/david/Plural/plural-cli/cmd/plural/deploy.go:118 +0x105
github.com/urfave/cli.HandleAction({0x1b330c0, 0x1fece40}, 0x5)
        /home/david/go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:524 +0xa8
github.com/urfave/cli.Command.Run({{0x1effa1f, 0x5}, {0x0, 0x0}, {0xc000702530, 0x1, 0x1}, {0x1f20a12, 0x15}, {0x0, ...}, ...}, ...)
        /home/david/go/pkg/mod/github.com/urfave/cli@v1.22.5/command.go:173 +0x652
github.com/urfave/cli.(*App).Run(0xc0001788c0, {0xc00013a000, 0x4, 0x4})
        /home/david/go/pkg/mod/github.com/urfave/cli@v1.22.5/app.go:277 +0x705
main.main()
        /home/david/Plural/plural-cli/cmd/plural/plural.go:346 +0x28f7
```

## Additional Info about Your Environment

---
#### Message from the maintainers:
Impacted by this bug? Give it a 👍. We factor engagement into prioritization.

PR #45, opened 2022-04-08
add error when building an app that isn't installed

DavidSpek:
## Summary
solves: https://github.com/pluralsh/plural-cli/issues/44
Adds an error message if a user tries to build an application that isn't installed for their user.


cmd/plural/deploy.go
```diff
@@ -113,6 +113,8 @@ func build(c *cli.Context) error {
 		installation, err := client.GetInstallation(c.String("only"))
 		if err != nil {
 			return err
+		} else if installation == nil {
```
michaeljguarino, line 116: I'd actually expect this to return an error at the client level.  I'll look into whether there's not an api bug there.

michaeljguarino (approved):

PR #47, opened 2022-04-14
Windows cli updates

dustinfaulkner91:
## Summary

**Updated files for windows file paths.
Added helper function to sanitize filepaths for windows.
some minor code cleanup.
Requires testing on non-windows OS.**
solves issues with windows reading slashes as escape characters and the cli failing most commands.

bug-fix

NA

## Test Plan

manual QA:
commands:
- version [x]
- build [x]
- deploy [x]
- diff [x]
- bounce [x]
- destroy
- init [x]
- bundle [x]
- link [x]
- unlink [x]
- help [x]

- repos [x]
- api [x]
- upgrade [/]

- watch 
- info [x]
- proxy [x]
- logs [x]
- ops [x]

- utils

- apply
- test
- push
- template
- from-grafana

- login [x]
- import [x]
- crypto [x]
- config [x]
- profile [x]

- create [x]

- validate [x]
- topsort [x]
- serve [x]
- shell
- workspace
- output
- build-context
- changed [x]

- help

## Checklist


pkg/utils/pathing/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+	switch os {
+	case "windows":
+		return strings.Replace(filepath, "\\", "/", -1)
+	default:
+		return filepath
+	}
+}
+
+// returns a string (filepath) with appropriate slashes, dependent on OS if error is empty. Optional parameters and method overloading unavailable in Go.
+func SanitizeFilePathWithError(filepath string, err error) (string, error) {
```
michaeljguarino, line 20: What would this function be used for out of curiosity?

michaeljguarino (commented):
Thanks for the contribution!  I think this generally looks good, just one q

pkg/wkspace/builder.go
```diff
@@ -218,6 +223,6 @@ func mkdir(path string) error {
 }
 
 func manifestPath(repo string) string {
-	path, _ := filepath.Abs(filepath.Join(repo, "manifest.yaml"))
+	path, _ := filepath.Abs(pathing.SanitizeFilepath(filepath.Join(repo, "manifest.yaml")))
```
DavidSpek, line 226: Same here, `filepath.Abs()` might be undoing the changes that `pathing.SanitizeFilepath()` does.

pkg/utils/tar.go
```diff
@@ -99,7 +101,7 @@ func untar(r io.Reader, dir, relpath string) (err error) {
 		if err != nil {
 			return err
 		}
-		abs := filepath.Join(dir, rel)
+		abs := pathing.SanitizeFilepath(filepath.Join(dir, rel))
 
 		fi := f.FileInfo()
 		mode := fi.Mode()
```
DavidSpek, line 107: I think the 2 functions `validRelativeDir` and `validRelPath` and the end of this file might also be relevant for Windows support.

DavidSpek (commented):
Thanks for the contribution! Here are a couple comments I've noticed while going through the code. During a test deployment on Linux I noticed there's still a debug log line somewhere that I'll try and find so it can be removed. Other than that nothing has broken on Linux, but I'll do further testing and also validate the changes on Windows.

cmd/plural/deploy.go
```diff
@@ -292,7 +294,7 @@ func handleDiff(c *cli.Context) error {
 	fmt.Printf("Diffing applications [%s] in topological order\n\n", strings.Join(sorted, ", "))
```
DavidSpek, line 294: This is one of the debug outputs that should be removed. Very strange that it isn't in the git diff though.

DavidSpek: Nevermind, I'm mistaken.

pkg/wkspace/builder.go
```diff
@@ -141,16 +142,16 @@ func (wk *Workspace) buildExecution(repoRoot string) error {
 	name := wk.Installation.Repository.Name
 	wkspaceRoot := filepath.Join(repoRoot, name)
 
-	if err := mkdir(filepath.Join(wkspaceRoot, ".plural")); err != nil {
+	if err := mkdir(pathing.SanitizeFilepath(pathing.SanitizeFilepath(filepath.Join(wkspaceRoot, ".plural")))); err != nil {
```
DavidSpek, line 145: Is there a reason for sanitizing the filepath twice here?

pkg/wkspace/builder.go
```diff
@@ -198,6 +200,9 @@ func DiffedRepos() ([]string, error) {
 		result[count] = repo
 		count++
 	}
+
+	fmt.Printf(" files %s    repos %s \n", files, repos)
```
DavidSpek, line 204: @dustinfaulkner91 This is one of the extra debug statements that should be removed.

pkg/utils/git/workspace.go
```diff
@@ -17,5 +18,7 @@ func Modified() ([]string, error) {
 			result = append(result, cols[1])
 		}
 	}
+
+	fmt.Printf(" res %s    result %s \n", res, result)
```
DavidSpek, line 22: @dustinfaulkner91 This is the other extra debug statement that should be removed.

pkg/utils/pathing/helpers.go
DavidSpek: There were some places where a function is called that returns a path and an error. So for those cases this could be used.

dustinfaulkner91:
I'll take a look when I get home.
I believe that function ended up not being used, so should have removed it.
If memory serves it was to handle the Abs functions that required a result and an error to be returned.

I'll see if the two mentioned functioned ( or others in that library) are a better solution for the minor Windows issues too.

Get Outlook for Android<https://aka.ms/AAb9ysg>
________________________________

DavidSpek:
Superseded by https://github.com/pluralsh/plural-cli/pull/54.

PR #51, opened 2022-04-28
Add Selects for Cloud Provider Regions

michaeljguarino:
## Summary
We currently require users to manually enter region names, which can be error prone, just use selects to solve for this instead.

## Test Plan
locally ran init for all clouds

## Checklist


pkg/provider/azure.go
```diff
@@ -31,6 +31,86 @@ type AzureProvider struct {
 	ctx           map[string]interface{}
 }
 
+var (
+	azureRegions = []string{
+		"East US",
```
DavidSpek, line 36: These are not the correct region codes for Azure.

DavidSpek: @michaeljguarino You can find the correct ones here. https://azuretracks.com/2021/04/current-azure-region-names-reference/

DavidSpek: Some of these option are most likely not valid for AKS. We should verify that this list (and maybe the other cloud providers as well) all support the managed Kubernetes services we are using.

michaeljguarino: I think they actually do still work when configured, I got them via: `az account list-locations | jq '.[].displayName'`

michaeljguarino: Azure seems to have some code to coerce them to the downcased/non-whitespaced version of the names

DavidSpek (approved):
Looks good to me!

PR #81, opened 2022-05-06
Fully Deprecate Helm Push

michaeljguarino:
## Summary

This provides a reimplementation of helm push's functionality straight into the plural cli, allowing us to both upload and download charts without direct helm or helm plugin interaction.

## Test Plan
Locally updated deps and push a new helm version

## Checklist


DavidSpek:
We should probably remove the helm-push installation from the Brew tap, see https://github.com/pluralsh/plural-cli/blob/master/.goreleaser.yaml#L109.

PR #89, opened 2022-05-16
Cli Github OAuth Setup

michaeljguarino:
## Summary

This uses githubs device flow to get gh creds in the cli and provision a repository automagically.  The advantage here is we can also auto-setup deploy keys ensuring proper git configuration for console installations at least for github users.

An extensible scm provider interface is defined as well for when we want to add gitlab+bitbucket support.

## Test Plan
Created a new repo locally using the new flow

## Checklist

Fixes PLU-32


pkg/scm/keys.go
```diff
@@ -0,0 +1,104 @@
...
+	found = true
+	return
+}
+
+func saveKeys(pub, priv string) error {
+	if !utils.Confirm("Would you like to save the keys to ~/.ssh?") {
+		return nil
+	}
+
+	pub, priv, err := keyFiles()
```
avaidyanatha, line 74: I might be misreading this but it looks like you're getting the filepath strings from `keyFiles()` here and saving the path as the byte array in the file instead of the generated key. Which would explain why the files generated for me were just path names.

Screen Shot 2022-05-17 at 11 16 40 PM

It seems like the contents of the keys are just their own directory paths?

Screen Shot 2022-05-17 at 11 31 06 PM

avaidyanatha (commented):
Going through the workflow, here are my notes.

1. It's not clear to me that I should run init without a repository created. Currently the help information says `initializes plural within a git repo`, so we should probably update the help description.

Screen Shot 2022-05-17 at 10 29 58 PM

2. If I say no to initializing a repo, I get a somewhat unhelpful error.

Screen Shot 2022-05-17 at 10 35 57 PM

I feel like I should just be told: "not inside a git repository" or something to that end, since it does seem like we do check for this in `handleInit`

*Selecting SCM went great, OAuth in the web browser was great, selecting an org was fine, then I hit this:*

3. the plural-cli was creating the private key with 644 permissions

Screen Shot 2022-05-17 at 10 40 39 PM 

To get this to work I had to manually chmod id_plural to 600 and then it let me proceed (presumably since it detected the existence of the keys it created already).

pkg/scm/keys.go
michaeljguarino: Oh shit this might all be buggy, I didn't test it thoroughly

PR #98, opened 2022-05-30
upgrade to aws-sdk-go-v2

DavidSpek:
## Summary
This PR migrates to `aws-sdk-go-v2`, which is necessary for us to automatically pickup temporary credentials created when logging in using `aws sso login`. For reference, these temporary credentials are stored in `~/.aws/cli/cache/*.json`.

## Test Plan
Create a new repository and run `plural init` and install an application, then run `plural build`. The CLI will try to create an S3 bucket for the terraform state.

## Checklist


pkg/provider/aws.go
```diff
@@ -204,18 +212,20 @@ func (aws *AWSProvider) Flush() error {
 }
 
 func (prov *AWSProvider) Decommision(node *v1.Node) error {
-	sess, err := session.NewSession(&aws.Config{
-		Region: aws.String(prov.Region()),
-	})
+	cfg, err := awsConfig.LoadDefaultConfig(context.TODO())
```
michaeljguarino, line 215: Let's set `ctx := context.TODO()` and pass them to all the funcs here instead of calling it again and again

DavidSpek: Then it's probably better to use `context.Background()` and add it to the provider struct.

pkg/provider/aws.go
```diff
@@ -23,8 +25,9 @@ type AWSProvider struct {
 	project       string
 	bucket        string
 	Reg           string `survey:"region"`
-	storageClient *s3.S3
+	storageClient *s3.Client
 	writer        manifest.Writer
+	goContext     context.Context
```
michaeljguarino, line 30: Oh I do think we can recreate these in individual funcs instead of storing in the provider object, which can be clunky in its own right.

DavidSpek: Without doing this, `mkBucket()` and `Decommission()` would have an empty context rather than the one used while configuring the AWS client. Although to pass the context around properly I think it should probably be a pointer. It didn’t seem to matter when I was just using `context.TODO()` everywhere, but never know when it might come back to cause problems.

michaeljguarino: why would it be bad to just use `context.Background()` or `context.TODO()` there?

michaeljguarino (approved):

PR #108, opened 2022-07-04
Prevent reuse of domain names

zreigz:
## Summary
Prevent reuse of domain names

## Checklist

### DIFF
--- a/pkg/bundle/configuration.go
+++ b/pkg/bundle/configuration.go
@@ -5,10 +5,10 @@ import (
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/mitchellh/go-homedir"
 	"github.com/pluralsh/plural/pkg/api"
 	"github.com/pluralsh/plural/pkg/manifest"
 	"github.com/pluralsh/plural/pkg/utils"
-	homedir "github.com/mitchellh/go-homedir"
 )
 
 func evaluateCondition(ctx map[string]interface{}, cond *api.Condition) bool {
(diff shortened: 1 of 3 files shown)


DavidSpek:
Is this for preventing the reuse of domain names between apps, within an account or within Plural globally?

pkg/manifest/context.go
```diff
@@ -25,7 +27,8 @@ type SMTP struct {
 
 type Context struct {
 	Bundles       []*Bundle
-	Buckets       []string
+	Buckets       sets.String
+	Domains       sets.String
```
DavidSpek, line 31: Changing this from a `[]string` to a map is causing errors with existing installations. We should use `[]string` for these, since a map with empty values might confuse users and be more error prone.

pkg/manifest/context.go
```diff
@@ -99,7 +100,21 @@ func (c *Context) HasBucket(bucket string) bool {
...
-} 
+}
+
+func (c *Context) AddDomain(bucket string) {
+	c.Domains = append(c.Domains, bucket)
+}
+
+func (c *Context) HasDomain(domain string) bool {
+	for _, d := range c.Domains {
+		if d == domain {
```
DavidSpek, line 111: can you add an exclusion here for `""`, there are some cases where an empty string for the hostname is used and we wouldn't want that to fail the uniqueness check.

zreigz: I've added this

pkg/bundle/configuration.go
```diff
@@ -60,36 +62,37 @@ func configure(ctx map[string]interface{}, item *api.ConfigurationItem, context
...
 		err = survey.AskOne(prompt, &res, opts...)
 		ctx[item.Name] = res
 	case Domain:
 		var res string
 		def = prevDefault(ctx, item, def)
-		prompt, opts := domainSurvey(def, item, proj)
+		prompt, opts := domainSurvey(def, item, proj, context)
 		err = survey.AskOne(prompt, &res, opts...)
 		ctx[item.Name] = res
+		context.AddDomain(res)
```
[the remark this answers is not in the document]
zreigz: done

PR #109, opened 2022-07-05
ENG-36 - Add bundle name column to the repos list command

floreks:
## Summary

Added a new column to the output of `plural repos list` command.

image

## Test Plan

Run the command locally

## Checklist

### DIFF
--- a/cmd/plural/repos.go
+++ b/cmd/plural/repos.go
@@ -2,9 +2,13 @@ package main
 
 import (
 	"fmt"
+	"strings"
+
+	"github.com/urfave/cli"
+
 	"github.com/pluralsh/plural/pkg/api"
 	"github.com/pluralsh/plural/pkg/format"
-	"github.com/urfave/cli"	
+	"github.com/pluralsh/plural/pkg/utils"
 )
 
 func reposCommands() []cli.Command {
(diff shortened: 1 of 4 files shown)


pkg/utils/map.go
```diff
@@ -57,13 +57,27 @@ func cleanUpMapValue(v interface{}) interface{} {
 
 func Dedupe(l []string) []string {
 	res := make([]string, 0)
-	seen := make(map[string]bool)
+	seen := make(map[string]struct{})
```
floreks, line 60: This is just a small optimization since the size of `struct{}` is 0 bytes.

floreks: Good idea, I'll update the method to use it.

cmd/plural/repos.go
```diff
@@ -54,14 +58,18 @@ func handleListRepositories(c *cli.Context) error {
 	addIcon := c.String("format") == "csv"
 
 	formatter := format.New(format.FormatType(c.String("format")))
-	header := []string{"Repo", "Description", "Publisher"}
+	header := []string{"Repo", "Description", "Publisher", "Bundle Name"}
```
michaeljguarino, line 61: Should the header be `Bundles` or `Bundle Names` (since there are multiple)

pkg/utils/map.go
```diff
@@ -56,14 +58,18 @@ func cleanUpMapValue(v interface{}) interface{} {
...
-		res = append(res, val)
-		seen[val] = true
+	return sets.NewString(l...).List()
+}
+
+type SimpleType interface {
+	string | int
+}
+
+func Map[T any, R SimpleType](slice []T, mapper func(elem T) R) []R {
```
michaeljguarino, line 68: We should make sure goreleaser is configured for the latest golang version to handle this

cmd/plural/repos.go
DavidSpek: You beat me to it haha

pkg/utils/map.go
DavidSpek: I'm pretty sure it is already.

floreks: I see that `go.mod` was also updated to use Go `1.18` already, that's why I wanted to try the new generics thingy :slightly_smiling_face:

michaeljguarino: Oh awesome

michaeljguarino (approved):

PR #111, opened 2022-07-06
Introduce linter action

zreigz:
## Summary
Fix all linter errors. Add linter GitHub action

## Checklist


cmd/plural/config.go
```diff
@@ -64,7 +64,7 @@ func handleAmend(c *cli.Context) error {
 	return config.Amend(c.Args().Get(0), c.Args().Get(1))
 }
 
-func handleRead(c *cli.Context) error {
+func handleRead(*cli.Context) error {
```
DavidSpek, line 67: this change looks a bit strange, especially given that the `handleConfigImport` function below looks essentially the same and doesn't have this change.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,28 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+      - main
```
maciaszczykm, line 8: This one is not needed as this repo is still using old naming.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,28 @@
+name: golangci-lint
+on:
+  push:
+    tags:
+      - v*
```
maciaszczykm, line 5: I think we may skip linters on tag branches. This is required in PRs and may be on master in case of direct commits.

cmd/plural/config.go
zreigz: yes you are right, linter doesn't complain

floreks: I guess both of them should have the variable name removed since they do not use context. It's strange that the `handleConfigImport` was not picked up by the linter.

.github/workflows/golangci-lint.yml
```diff
@@ -0,0 +1,25 @@
...
+      - uses: actions/setup-go@v3
+        with:
+          go-version: 1.18
+      - uses: actions/checkout@v3
+      - name: golangci-lint
+        uses: golangci/golangci-lint-action@v3
+        with:
+          # Optional: version of golangci-lint to use in form of v1.2 or v1.2.3 or `latest` to use the latest version
+          version: v1.46.2
+          args: --timeout 5m --verbose
```
floreks, line 25: Instead of using args, we could create a `.golangci.yaml` file to store the configuration and then it would be reused when running it both locally and on the CI. This way we would not have to keep it in sync.

https://golangci-lint.run/usage/configuration/#config-file

cmd/plural/config.go
zreigz: generally, the parameter is required to fulfill interface requirements. Linter suggests using just the type.

floreks:
I'd also enable more linter as by default there are only a couple of them used.
https://golangci-lint.run/usage/linters/

Most important I guess would be:
- `gofmt` - formatter
- `goimports` - formatter
- `gocyclo` - checks the cyclomatic complexity of the code
- `importas` - makes sure imports in all files use the same structure
- `goconst` - finds repeated strings that could be a const
- `misspell` - checks comments for common misspells
- `nestif` - checks if there are too many nested ifs
-

.golangci.yml
```diff
@@ -12,4 +12,7 @@ linters:
     - ineffassign
     - misspell
     - structcheck
```
DavidSpek, line 14: this should be removed since it isn't compatible with Go 1.18

cmd/plural/crypto.go
```diff
@@ -121,10 +121,13 @@ func cryptoCommands() []cli.Command {
 	}
 }
 
-func handleEncrypt(c *cli.Context) error {
+func handleEncrypt(*cli.Context) error {
```
DavidSpek, line 124: when I run the linter locally I'm not getting this change for some reason.

DavidSpek: Nevermind, issue with my setup.

DavidSpek: @zreigz Actually it does seem like this change isn't picked up when I run it locally. What command are you using to run `golangci-lint`?

zreigz: I had this when I was using go 1.17, now I changed to 1.18 and now it's not valid. I will revert

cmd/plural/validation.go
```diff
@@ -49,15 +49,15 @@ func owned(fn func(*cli.Context) error) func(*cli.Context) error {
 	}
 }
 
-func confirmed(fn func(*cli.Context) error, msg string) func(*cli.Context) error {
-	return func(c *cli.Context) error {
-		if ok := confirm(msg); !ok {
-			return nil
-		}
+// func confirmed(fn func(*cli.Context) error, msg string) func(*cli.Context) error {
```
michaeljguarino, line 52: is this really unused?  If it is, we can just eliminate it.  I know i used it previously

pkg/utils/validation.go
```diff
@@ -12,35 +12,35 @@ const (
...
-			continue
-		}
-
-		return res
-	}
-}
+// type fetcher func() (string, error)
+// type validator func(string) error
+
+// func UntilValid(fetch func() error) {
```
michaeljguarino, line 18: rm

michaeljguarino (commented):
In general, I'm not sure I love having to handle every error since sometimes it truly isn't necessary and clunks up the code.

Also I'm not sure all the commented code is actually unused, can you confirm that's the case?

DavidSpek (approved):
Thanks for getting the CLI code cleaned up, it's been long overdue!

PR #117, opened 2022-07-14
return only lock object when lock file doesn't exist

zreigz:
## Summary
This PR fixes the` plural apply` command. The bug was introduced during linter fixes

## Checklist


DavidSpek (approved):
Seems legit, thanks for the fix!

Issue #105, opened 2022-06-30
feat: Add check for onplural.sh domain entry


DavidSpek:
We should limit people to only be able to create a subdomain under `onplural.sh` that's a single level deep (`xxx.onplural.sh` and not `xxx.yyy.onplural.sh`). The second option will cause an error in our external-dns deployment where graphql can't find the domain, even though it is created in Plural.

PR #165, opened 2022-08-03
Add check for onplural.sh domain entry

zreigz:
## Summary
Check single-level deep domain
Fixes: #105

## Checklist


pkg/utils/validation.go
```diff
@@ -12,6 +12,21 @@ const (
 	dnsRegex = "(([a-zA-Z0-9]|[a-zA-Z0-9][a-zA-Z0-9\\-]*[a-zA-Z0-9])\\.)*([A-Za-z0-9]|[A-Za-z0-9][A-Za-z0-9\\-]*[A-Za-z0-9])"
 )
 
+func ValidateSingleLevelDeep(val, subdomain string) error {
```
michaeljguarino, line 15: We should actually implement this in the api as well.

michaeljguarino (approved):
I've merged the api-level check, but we can keep the check here too

PR #170, opened 2022-08-05
ENG-457 - Refactor plural serve and improve logging

floreks:
## Summary

- Added better error handling to the `setupCli`
- Improve overall logging messages and AWS exec command handling to return an actual error message instead of exit status code.
- Added `make serve` target to build and run local cloud cli container with `plural serve`. It exposes the server on the `localhost:8080` and makes the `v1/setup` endpoint easily accessible for tests.

This PR improves overall logic and logging, but does not fully solve the main issue. We should also make sure that the response error is correctly propagated all the way to the frontend.

#### Server log
image

#### Server response
image

## Test Plan

1. `make serve`
2. POST to `http://localhost:8080/v1/setup` with the correct setup request structure

## Checklist


pkg/server/setup.go
```diff
@@ -93,6 +94,10 @@ func setupCli(c *gin.Context) error {
 		return err
 	}
 
+	if err := setupProvider(&setup); err != nil {
```
michaeljguarino, line 97: I'm pretty sure reordering these will break the ability to clone the git repo (since the directory will already exist)

floreks: Correct me if I am wrong, but it looks like the `setupProvider` step does not directly touch the `~/workspace` directory.
For the GCP it writes the credentials to the `~/gcp.json` and for AWS it uses `aws configure set ...` command to alter config files in the `~/.aws` directory. It should not have an impact on the `setupGit` step and `~/workspace` directory creation during the `git clone`.

michaeljguarino: Oh you're right, this makes sense

michaeljguarino (approved):

PR #172, opened 2022-08-08
ENG-431 plural deploy resiliency

zreigz:
## Summary
Add additional retries to TF and helm (1 more retry + a sleep)
Change message at the end of 'plural deploy'

## Checklist

### DIFF
--- a/cmd/plural/deploy.go
+++ b/cmd/plural/deploy.go
@@ -1,10 +1,12 @@
 package main
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/pluralsh/plural/pkg/api"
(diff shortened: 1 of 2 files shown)


DavidSpek:
@zreigz We actually already have a place where retries are being set for Terraform and Helm. It's in the `deploy.hcl` files. See [here](https://github.com/pluralsh/plural-cli/blob/master/pkg/executor/default.go#L9)

zreigz:
@DavidSpek I've added this retry according to the issue description.  Should I remove this and only change the info message?

DavidSpek:
Yeah I think remove that retry, but leave change to the info message. Then you can change https://github.com/pluralsh/plural-cli/blob/master/pkg/executor/default.go#L29 and https://github.com/pluralsh/plural-cli/blob/master/pkg/executor/default.go#L62 to `2`.

zreigz:
@DavidSpek PTAL

DavidSpek (approved):
Looks good, thanks for the changes!

PR #174, opened 2022-08-08
Update `plural version` command

maciaszczykm:
## Summary
- Fix `plural version` command to display compilation time set in ldflags
  - Use RFC3339 date format and UTC time zone
  - Use same names as in [goreleaser defaults](https://goreleaser.com/cookbooks/using-main.version)
  - Update format to match `plural` command (see below)
- Update build command in Dockerfile
- Restore recency check that was removed in https://github.com/pluralsh/plural-cli/commit/a017714500b929639c56e436a64d794528a1262d as [Homebrew publish is already automated](https://github.com/pluralsh/homebrew-plural/commits/main) (closes ENG-452)
  - Base check on version numbers instead of commit SHA
  - It will run only when using `plural version` command, so it will not show too often. Previously this was based on random numbers with 25% chance of warning being displayed. Would you like to restore that behavior?
- Shorten release description

Zrzut ekranu 2022-08-9 o 10 22 26

## Test Plan
Tested manually.

## Checklist

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -7,7 +7,7 @@ builds:
   - env:
       - CGO_ENABLED=0
     ldflags:
-    - "-s -w -X main.Version={{.Version}} -X main.GitCommit={{.Commit}} -X github.com/pluralsh/plural/pkg/scm.GitlabClientSecret={{ .Env.GITLAB_CLIENT_SECRET }}"
+    - "-s -w -X main.version={{.Version}} -X main.commit={{.Commit}} -X main.date={{.Date}} -X github.com/pluralsh/plural/pkg/scm.GitlabClientSecret={{ .Env.GITLAB_CLIENT_SECRET }}"
     goos:
       - linux
       - windows
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -34,11 +34,12 @@ changelog:
       - '^test:'
 release:
   name_template: "{{.ProjectName}}-v{{.Version}}"
+  header: |
+    ## Plural CLI release ({{ .Date }})
+    Welcome to this new release of the Plural CLI!
   footer: |
     ## Thanks!
-
     We'd like to thank all contributors for helping with improving our CLI!
-
 brews:
   - name: plural
     tap:
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -34,10 +34,6 @@ changelog:
       - '^test:'
 release:
   name_template: "{{.ProjectName}}-v{{.Version}}"
-  header: |
-    ## Plural CLI release ({{ .Date }})
-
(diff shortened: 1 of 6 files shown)


DavidSpek:
just an FYI, the version is coming from the git tag which should be correct (except you might want to strip the `v` if you want same naming as the goreleaser).

.goreleaser.yaml
```diff
@@ -34,10 +34,6 @@ changelog:
       - '^test:'
 release:
   name_template: "{{.ProjectName}}-v{{.Version}}"
-  header: |
```
DavidSpek, line 37: Why is this removed? It's used to generate the GitHub release.

maciaszczykm: I thought that this was coming from boilerplate and removed it to make desc a bit shorter. Do you want to keep that?

maciaszczykm: Sure, done!

cmd/plural/version.go
```diff
@@ -7,23 +7,24 @@ import (
...
 
 	"github.com/pluralsh/plural/pkg/utils"
 	"github.com/urfave/cli"
 )
 
-var (
-	GitCommit string
-	Version   string
+const (
+	devVersion = "dev"
```
DavidSpek, line 16: I think the logic to detect a dev version isn't fully correct. I get this after running `make install`:

```console
david@david-pop-os:~/Plural/plural-cli$ plural version
PLURAL CLI:
   version      v0.4.4-74-ge462cd9
   git commit   e462cd9
   compiled at  2022-08-09T11:41:56+0000
   os/arch      linux/amd64

Your version appears out of date, try updating it with your package manager
```

We should probably have message for dev version telling people they are using a dev version and pointing them to where they can install official releases.

maciaszczykm: >I think the logic to detect a dev version isn't fully correct.

It shows `dev` if the build was done without any params, so without using any of the `make` commands. 

It was intended as I think in most of the cases we should avoid putting `dev` placeholder here and be as precise as possible. That's because most common use cases I see for this are checking for updates, debugging and reporting issues where every commit makes difference.

We can change that and show `dev` if the git distance is bigger than 0 but that would put all non-release versions into one bracket where only difference would be compilation time. In my opinion it would be better to avoid that and do the thing that you have suggested instead:

> We should probably have message for dev version telling people they are using a dev version and pointing them to where they can install official releases.

With this info we can warn people about using dev/non-official release but still keep all the detailed version info. What do you think?

DavidSpek: That sounds good to me.

maciaszczykm: It's done:

Zrzut ekranu 2022-08-10 o 09 16 33

michaeljguarino (approved):
awesome, let's merge it

PR #175, opened 2022-08-10
ENG-429 fix resource group creation for azure

zreigz:
## Summary
Create a resource group if doesn't exist for the Azure provider
Unit tests

## Checklist

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -3,27 +3,76 @@ package provider
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"net/http"
 	"net/url"
 	"os"
 	"os/exec"
 
-	"github.com/pluralsh/plural/pkg/kubernetes"
-
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/Azure/azure-sdk-for-go/sdk/azcore"
+	"github.com/Azure/azure-sdk-for-go/sdk/azidentity"
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
 	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
 	"github.com/Azure/azure-storage-blob-go/azblob"
 	"github.com/Azure/go-autorest/autorest"
+	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
 	"github.com/azure/azure-sdk-for-go/services/compute/mgmt/2021-07-01/compute"
 	"github.com/pluralsh/plural/pkg/config"
+	"github.com/pluralsh/plural/pkg/kubernetes"
 	"github.com/pluralsh/plural/pkg/manifest"
 	"github.com/pluralsh/plural/pkg/template"
 	"github.com/pluralsh/plural/pkg/utils"
-	"github.com/pluralsh/plural/pkg/utils/errors"
+	pluralerr "github.com/pluralsh/plural/pkg/utils/errors"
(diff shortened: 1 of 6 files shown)


pkg/provider/azure_test.go
```diff
@@ -0,0 +1,131 @@
...
+	"github.com/Azure/azure-sdk-for-go/sdk/resourcemanager/resources/armresources"
+	"github.com/Azure/azure-sdk-for-go/services/storage/mgmt/2019-06-01/storage"
+	"github.com/pluralsh/plural/pkg/manifest"
+	"github.com/pluralsh/plural/pkg/provider"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCreateResourceGroup(t *testing.T) {
+	resourceGroupName := "test"
+	tests := []struct {
```
michaeljguarino, line 15: could we not define these structs inline (can be separate private structs in this file)?  Will just reduce the line lengths of these tests and improve readability a bit I think

pkg/provider/azure.go
```diff
@@ -3,35 +3,85 @@ package provider
...
 	"github.com/pluralsh/plural/pkg/template"
 	"github.com/pluralsh/plural/pkg/utils"
-	"github.com/pluralsh/plural/pkg/utils/errors"
+	pluralerr "github.com/pluralsh/plural/pkg/utils/errors"
 	v1 "k8s.io/api/core/v1"
 )
 
+// ResourceGroupClient is the subset of functions we need from armresources.VirtualResourceGroupsClient;
+// this interface is purely here for allowing unit tests.
+type ResourceGroupClient interface {
```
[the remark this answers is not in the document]
zreigz: unfortunately not, azure's `armresources.ResourceGroupsClient` is a struct like the other clients

pkg/provider/azure_test.go
zreigz: When we have one test scenario in the file then we can define the struct with the test parameters:
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/cmd/convert/convert_test.go#L31

 But for multiple test functions, it's a common pattern to use an inline struct for the test parameters: https://github.com/kubernetes/kubernetes/blob/ea0764452222146c47ec826977f49d7001b0ea8c/pkg/kubelet/kuberuntime/helpers_linux_test.go

pkg/provider/azure.go
```diff
@@ -127,13 +181,26 @@ func mkAzure(conf config.Config) (prov *AzureProvider, err error) {
...
+	var err error
+	clients := clientSet
+	if clientSet == nil {
+		clients, err = GetClientSet(utils.ToString(man.Context["SubscriptionId"]))
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	return &AzureProvider{man.Cluster, man.Project, man.Bucket, man.Region, man.Context, nil, clients}, nil
```
DavidSpek, line 194: I'm not sure if we need to initialize the client here, since we also don't initialize the writer.

zreigz: we have to initialize the clients here for two reasons:
 - to initialize fake clients for the tests (AzureProvider has private fields and can be done in the same package) and it's the main reason
 - previously the clients were created on demand before the call, now after refactor they must exist when we initialize the Azure provider

DavidSpek: Now that I think about it I believe I did something similar with the AWS provider a while back.

pkg/provider/azure.go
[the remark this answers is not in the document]
michaeljguarino: Ugh

pkg/provider/azure_test.go
michaeljguarino: If it's common golang style then that's cool, just makes the func a bit longer

michaeljguarino (approved):

PR #178, opened 2022-08-11
Fix crypto error handling

michaeljguarino:
The error check was being executed prematurely, which breaks some console shared repos (where the filesystem is read-only and a crypto provider fails to materialize, but you can still use the age provider)

## Test Plan
ensure compiles

## Checklist


DavidSpek (approved):
This should solve the problem we were seeing.

PR #179, opened 2022-08-12
ENG-497 fix plural crypto recover command

zreigz:
## Summary
This PR fixes the` plural crypto recover` command. It will also fix the broken `~/.plural/key`

## Checklist

### DIFF
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -8,7 +8,6 @@ import (
 	"os"
 	"path/filepath"
 	"strconv"
-	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/mitchellh/go-homedir"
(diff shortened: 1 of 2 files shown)


cmd/plural/crypto.go
```diff
@@ -371,8 +372,14 @@ func handleRecover(c *cli.Context) error {
 	if !ok {
 		return fmt.Errorf("could not find `key` in console-conf secret")
 	}
+	keyValue := string(key)
```
DavidSpek, line 375: Wouldn't it be possible to reuse the AESKey struct and related functions here?

```go
type AESKey struct {
	Key string
}
```

They're in `pkg/crypto/key.go`.

zreigz: good point let me check this

zreigz: works fine, thanks

DavidSpek (approved):
Thanks for the quick response for this issue!

PR #207, opened 2022-08-29
fix crypto command

zreigz:
## Summary
The `requireArgs` function is not needed because name and email are passed as a flag but not an argument 

## Checklist

### DIFF
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -116,7 +116,7 @@ func (p *Plural) cryptoCommands() []cli.Command {
 					Usage: "a email to share with (multiple allowed)",
 				},
 			},
-			Action: requireArgs(p.handleCryptoShare, []string{"email"}),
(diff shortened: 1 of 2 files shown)


cmd/plural/crypto.go
```diff
@@ -127,7 +127,7 @@ func (p *Plural) cryptoCommands() []cli.Command {
 					Usage: "a name for the key",
 				},
 			},
-			Action: requireArgs(p.handleSetupKeys, []string{"name"}),
+			Action: p.handleSetupKeys,
```
michaeljguarino, line 130: Is name a flag there?

zreigz: yes, there is a `name` flag but the check was for the argument.

zreigz:
@michaeljguarino I fixed also the action function PTAL

michaeljguarino (approved):

PR #208, opened 2022-08-29
Make `crypto share` additive

michaeljguarino:
## Summary

We'd forcibly overwrite all identities when sharing previously, which can cause weird issues.  This should ensure that we only add to the share list.

## Test Plan
tested on the plural-az-installations repo

## Checklist

### DIFF
--- a/pkg/crypto/age.go
+++ b/pkg/crypto/age.go
@@ -117,9 +117,20 @@ func SetupAge(client api.Client, emails []string) error {
 			return err
 		}
 
+		dedupeKey := func(id *AgeIdentity) string { return fmt.Sprintf("%s::%s", id.Email, id.Key) }
+
 		idents := ageConfig.Identities
+		present := map[string]bool{}
+		for _, id := range idents {
+			present[dedupeKey(id)] = true
+		}
+
 		for _, key := range keys {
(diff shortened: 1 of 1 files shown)


pkg/crypto/age.go
```diff
@@ -117,9 +117,20 @@ func SetupAge(client api.Client, emails []string) error {
 			return err
 		}
 
+		dedupeKey := func(id *AgeIdentity) string { return fmt.Sprintf("%s::%s", id.Email, id.Key) }
+
 		idents := ageConfig.Identities
+		present := map[string]bool{}
```
zreigz, line 123: "k8s.io/apimachinery/pkg/util/sets"

present := sets.sets.NewString

zreigz:
```
		present := sets.NewString()
		for _, id := range idents {
			present.Insert(dedupeKey(id))
		}

		for _, key := range keys {
			id := &AgeIdentity{Key: key.Content, Email: key.User.Email}
			if !present.Has(dedupeKey(id)) {
				idents = append(idents, id)
			}
			
		}

```

michaeljguarino: oh cool, i was hoping someone would build this

michaeljguarino:
@zreigz used the sets package

PR #212, opened 2022-09-01
ENG-560 add global flags

zreigz:
## Summary
This PR adds two global flags: `profile-file` and `encryption-key-file`. User can also specify env variables:
`PLURAL_PROFILE_FILE` and `PLURAL_ENCRYPTION_KEY_FILE`

## Checklist

### DIFF
--- a/cmd/plural/api.go
+++ b/cmd/plural/api.go
@@ -63,6 +63,7 @@ func (p *Plural) apiCommands() []cli.Command {
 }
 
 func (p *Plural) handleInstallations(c *cli.Context) error {
+	p.InitPluralClient()
 	installations, err := p.GetInstallations()
 	if err != nil {
 		return err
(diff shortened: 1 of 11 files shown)


cmd/plural/api.go
```diff
@@ -63,6 +63,7 @@ func (p *Plural) apiCommands() []cli.Command {
 }
 
 func (p *Plural) handleInstallations(c *cli.Context) error {
+	p.InitPluralClient()
```
DavidSpek, line 66: Why do we need to init the plural client everywhere now?

DavidSpek (approved):
Tested it and works as it should. Thanks for implementing this!

cmd/plural/api.go
zreigz: because parameters and env variables are taken in the runtime but the plural client needs the profile file before. In this case, the client should be initialized during command execution when the parameters are fetched

PR #226, opened 2022-09-19
Make plural repair fail silently

michaeljguarino:
## Summary
The console can sometimes become jammed on unusual encrpytion diffs, but they can be fixed by just adding and committing all files and pushing upstream.  `plural repair` should do that, but it's not safe to use in the console if it blocks deploys, so having it always exit 0 is the safest bet.

## Test Plan
run it locally

## Checklist

### DIFF
--- a/cmd/plural/git.go
+++ b/cmd/plural/git.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"fmt"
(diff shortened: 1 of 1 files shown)


zreigz (approved):
lgtm

PR #236, opened 2022-10-12
Begin working on azure cloud shell support

michaeljguarino:
## Summary

It looks like the only effective way to provide local creds is via env vars.  It's a little roundabout since the terminal session needs the env vars not the server proc, so add a .env file and source it at terminal mount

## Test Plan
will need to end-to-end test this once available

## Checklist


zreigz (approved):
lgtm

PR #248, opened 2022-11-01
Add ability to only deploy app + dependencies

michaeljguarino:
## Summary

This pr introduces two things:

* `plural dependencies REPO` - lists deps for a repo in topo order
* `--from` flag for `plural deploy` - this will deploy only an app and its deps in appropriate order

I also added a fair number of go generics utils that were helpful here and likely in the future.  We should probably create a separate lib for those soon.

## Test Plan
tested locally

## Checklist


zreigz (approved):
lgtm

PR #250, opened 2022-11-02
Use generics to simplify table printing

michaeljguarino:
## Summary
Defines a PrintTable[T] function to eliminate a ton of boilerplate table printing code

## Test Plan
local

## Checklist

### DIFF
--- a/cmd/plural/api.go
+++ b/cmd/plural/api.go
@@ -1,9 +1,9 @@
 package main
 
 import (
-	"os"
-
-	"github.com/olekukonko/tablewriter"
+	"github.com/pluralsh/plural/pkg/api"
+	"github.com/pluralsh/plural/pkg/utils"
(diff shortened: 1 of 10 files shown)


DavidSpek:
@michaeljguarino Note that we'll need to make sure the dockerfiles are updated to the latest Go version so that generics are supported when building.

zreigz:
> @zreigz I'm not entirely sure why the unit test is failing here, any idea? It looks like some odd interaction with the git checkout code

I will check this

zreigz:
@michaeljguarino this error is from the system call. Every call of this command `os.Getwd()` in the failing test causes the error. For now, I skipped the test

zreigz:
@michaeljguarino found the problem, I've pushed the fix

zreigz (approved):
lgtm

PR #252, opened 2022-11-03
ENG-756 check for newest CLI

zreigz:
## Summary
Display a warning when the new CLI version is available.

## Checklist


michaeljguarino (commented):
We can do this, but we should **NOT** do it on any crypto commands, as it will corrupt the repo

A previous cli version did this and hit that issue and we quickly reverted.

zreigz:
> We can do this, but we should **NOT** do it on any crypto commands, as it will corrupt the repo
> 
> A previous cli version did this and hit that issue and we quickly reverted.

done

DavidSpek:
@zreigz Can you maybe add comments in the code where the crypto commands are so that we don't accidentally make that mistake again in the future.

pkg/utils/version.go
```diff
@@ -0,0 +1,28 @@
+package utils
+
+import (
+	"context"
+
+	semverlib "github.com/Masterminds/semver/v3"
+	"github.com/google/go-github/v45/github"
+)
+
+func CheckLatestVersion(currentVersion string) {
```
maciaszczykm, line 10: There is a check for latest version in the `plural version` command already, see `cmd/plural/version.go`.

maciaszczykm: Zrzut ekranu 2022-11-3 o 14 57 52

zreigz: I've added this check as awrapper for all CLI commands

pkg/utils/version.go
```diff
@@ -0,0 +1,29 @@
...
+	client := github.NewClient(nil)
+	latestRelease, _, err := client.Repositories.GetLatestRelease(context.Background(), "pluralsh", "plural-cli")
```
maciaszczykm, line 12: This will returned latest release looking at date not version, see https://docs.github.com/en/rest/releases/releases#get-the-latest-release.

maciaszczykm: If we would create any patch release then it would not work.

zreigz: I also get tag release and it returns the version, exactly what you did

maciaszczykm: We'll need to order it by version not date to fix that.

zreigz: @maciaszczykm you are right. I will fix it. Thanks

pkg/utils/version.go
```diff
@@ -0,0 +1,29 @@
...
+		latestVersion := *latestRelease.TagName
+		cv, err := semverlib.NewVersion(currentVersion)
+		if err != nil {
+			return
+		}
+		lv, err := semverlib.NewVersion(latestVersion)
+		if err != nil {
+			return
+		}
+		if cv.Major() == lv.Major() && cv.Minor() == lv.Minor() && cv.Patch() >= lv.Patch() {
```
michaeljguarino, line 23: Could we just not check the patch version at all?  Trying to think of any case where it'd be ahead of the gh version, but seems like basically equivalent with a bit more coverage

zreigz:
@maciaszczykm PTAL

pkg/utils/version.go
zreigz: the patch version is the last number from version v0.5.20 -> 20 so it does matter

michaeljguarino (approved):

cmd/plural/crypto.go
```diff
@@ -51,6 +51,8 @@ const Gitignore = `/**/.terraform
 .vscode
 `
 
+// IMPORTANT
+// Don't wrap Action methods
```
michaeljguarino, line 55: If we want to put the full explanation, it's basically:

repo cryptography relies on git smudge and clean filters, which pipe a file into stdin and respond with a new version of the file from stdout.  If we write anything besides the crypto text, it will no longer be decryptable naturally.

pkg/utils/version.go
michaeljguarino: oh wait I read this backwards, this makes more sense now

michaeljguarino (changes requested):
Actually, wait, we should coinflip the CheckLatestVersion function

cmd/plural/validation.go
```diff
@@ -141,3 +142,10 @@ func repoRoot() error {
 
 	return nil
 }
+
+func latestVersion(fn func(*cli.Context) error) func(*cli.Context) error {
+	return func(c *cli.Context) error {
+		utils.CheckLatestVersion(version)
```
michaeljguarino, line 148: We should run this maybe 1 in 10 or 1 in 5 executions, and also ignore entirely when run in the console (there are other cases where we do this)

zreigz: @michaeljguarino 
For random execution: 
 1) should I add some random trigger inside the function?
```
	rand.Seed(time.Now().UnixNano())
	if rand.Intn(5) == 4 {
		utils.CheckLatestVersion(version)
	}

```
 2) save the number of execution in ~/.plural/config.yml to make the decision on when to enable the check


For the console:
1) add CLI global flag with the env variable (https://github.com/pluralsh/plural-cli/blob/master/cmd/plural/plural.go#L418) to disable the `utils.CheckLatestVersion(version)` method
for example:
```
$ PLURAL_DISABLE_LATEST_CHECK=true plural init
```
by default will be false for regular usage.

2) something else maybe?

michaeljguarino: Yeah a random coinflip like that looks perfect.  We should probably add a coinflip function to polly too since it's generally very useful.

I think we should just have a general env var that tells the cli it's in the console, could literally be `PLURAL_CONSOLE=1` or something

zreigz: @michaeljguarino thanks, I will follow your suggestions

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):

PR #254, opened 2022-11-04
Fix bundle list command

maciaszczykm:
## Summary
`plural bundle list ...` was failing outside repositories as `provider` param was undefined. Is was not possible to set it to `nil` as it's `string` type, so new query was added to our client.

Closes ENG-639.

Zrzut ekranu 2022-11-4 o 09 47 52

## Test Plan
Tested manually.

## Checklist


zreigz (approved):
lgtm

PR #255, opened 2022-11-04
ENG-648 add error if sharing with user that has no keys setup

zreigz:
## Summary
Currently, if a user tries to share a repo with somebody that doesn't have keys setup using plural crypto setup-key it fails silently. This PR fixes this issue.

## Checklist

### DIFF
--- a/pkg/crypto/age.go
+++ b/pkg/crypto/age.go
@@ -15,6 +15,8 @@ import (
 	"github.com/pluralsh/plural/pkg/utils"
 	"github.com/pluralsh/plural/pkg/utils/git"
 	"github.com/pluralsh/plural/pkg/utils/pathing"
+	"github.com/pluralsh/polly/algorithms"
+	"github.com/pluralsh/polly/containers"
 	"gopkg.in/yaml.v2"
 	"k8s.io/apimachinery/pkg/util/sets"
 )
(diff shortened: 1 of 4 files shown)


pkg/crypto/age.go
```diff
@@ -104,6 +104,20 @@ func Identity() (*age.X25519Identity, error) {
...
+		return emails
+	}
+	emailSet := sets.NewString(emails...)
+	emailKeySet := sets.NewString()
+	for _, key := range keys {
+		if key.User != nil && key.User.Email != "" {
+			emailKeySet.Insert(key.User.Email)
+		}
+	}
+	return emailSet.Difference(emailKeySet).List()
```
michaeljguarino, line 118: Is there no empty function to call here and have this func be a bool?

Can also use the new set I implemented in our go lib, which has a set equality func

pkg/crypto/age.go
```diff
@@ -104,6 +104,20 @@ func Identity() (*age.X25519Identity, error) {
 	return generateIdentity(getAgePath())
 }
 
+func findMissingKeyForEmail(emails []string, keys []*api.PublicKey) []string {
+	if len(keys) == 0 {
+		return emails
+	}
+	emailSet := sets.NewString(emails...)
+	emailKeySet := sets.NewString()
+	for _, key := range keys {
```
michaeljguarino, line 113: Would be more elegant to use a Map func here

pkg/crypto/age.go
zreigz: done

pkg/crypto/age.go
zreigz: done

zreigz:
@michaeljguarino PTAL

pkg/crypto/age.go
```diff
@@ -116,6 +134,10 @@ func SetupAge(client api.Client, emails []string) error {
 		if err != nil {
 			return err
 		}
+		missingEmails := findMissingKeyForEmail(emails, keys)
```
michaeljguarino, line 137: findMissingKeyForEmail should just return the set and then call empty below (saves an array allocation and should be constant time)

zreigz: yes but when I want to display emails that have no keys setup I will have this:
```
Some of the users map[test@email.com:true] have no keys setup
```
instead of this message:
```
Some of the users [test@email.com] have no keys setup
```
So in the current version, I don't need to make any conversion

michaeljguarino (approved):

PR #281, opened 2022-11-17
ENG-965 improve plural destroy

zreigz:
## Summary
Auto-delete DNS records on full cluster teardown
Fix unit tests

## Checklist


cmd/plural/deploy.go
```diff
@@ -386,7 +386,15 @@ func (p *Plural) destroy(c *cli.Context) error {
 			return api.GetErrorResponse(err, "GetInstallation")
 		}
 
-		return p.doDestroy(repoRoot, installation)
+		if err := p.doDestroy(repoRoot, installation); err != nil {
+			return err
+		}
+		if confirm(fmt.Sprintf("Would you like to remove %s from app.plural.sh?", repoName)) {
+			if err := p.ResetInstallation(installation.Id); err != nil {
```
michaeljguarino, line 393: I actually have a change that does this as well that might be a bit more in line with what I'd expect the right tear-down is, could we perhaps rip out the uninstall code and limit this to just calling destroyCluster at the end

cmd/plural/deploy.go
```diff
@@ -428,6 +436,21 @@ func (p *Plural) destroy(c *cli.Context) error {
...
 
+	if confirm("Would you like to remove installations from app.plural.sh?") {
+		if err := p.handleResetInstallations(c); err != nil {
+			return err
+		}
+		man, err := manifest.FetchProject()
+		if err != nil {
+			return err
+		}
+		if err := p.DestroyCluster(man.Network.Subdomain, man.Cluster, strings.ToUpper(man.Provider)); err != nil {
```
michaeljguarino, line 447: We can actually just always call this api w/o confirmation, it should be safe.

That said, we should only do it when it's not a single app destroy (so check `repoName != ""`)

maciaszczykm (approved):

PR #287, opened 2022-11-30
[WIP] add lua template

zreigz:
## Summary
This PR introduces support for Lua scripts. The user must define the global table variable: `output = {}` which represents all YAML values. All variables are passed through the `Var`, for example: `Var.Values.git_email`.
```
output = {
    provider=Var.Provider,
    ingress={
            annotations={
                "kubernetes.io/tls-acme: 'true'"
             }
 }

 if Var.Provider == "azure" then
        output.ingress.annotations = {
            "external-dns.alpha.kubernetes.io/target: '127.0.0.1'"
        }
 end
 
```

## Checklist


pkg/scaffold/helm.go
```diff
@@ -234,15 +214,50 @@ func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 	values["plrl"] = map[string]interface{}{
 		"license": w.Installation.LicenseKey,
 	}
+	return values, nil
+}
 
-	io, err := yaml.Marshal(values)
+func fromGoTemplate(vals map[string]interface{}, globals map[string]interface{}, values map[string]map[string]interface{}, w *wkspace.Workspace, chartInst *api.ChartInstallation) error {
```
michaeljguarino, line 220: should we perhaps put these template implementations into a subpackage w/in scaffold?

michaeljguarino (commented):
I think we should have the name of the output variable be slightly less clunky.  Actually just naming it `output` might be more than sufficient

pkg/scaffold/template/lua.go
```diff
@@ -0,0 +1,56 @@
...
+		L.SetGlobal(name, luar.New(L, function))
+	}
+	for name, function := range sprig.GenericFuncMap() {
+		L.SetGlobal(name, luar.New(L, function))
+	}
+
+	if err := L.DoString(tplate); err != nil {
+		return nil, err
+	}
+	if err := utils.MapLua(L.GetGlobal("valuesYaml").(*lua.LTable), &output); err != nil {
```
michaeljguarino, line 29: I still think the name `valuesYaml` is a bit clunky.  Feel like something more straightforward like `output` would look a bit better and be nicer for external devs to use as well in the future.

michaeljguarino (approved):

PR #292, opened 2022-12-13
ENG-4 add bitbucket

zreigz:
## Summary
Add support for bitbucket

## Checklist


pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return Context{}, err
+	}
+
+	wl, err := b.Client.Workspaces.List()
+	if err != nil {
+		return Context{}, err
+	}
+
+	workspaces := make([]string, 0)
```
michaeljguarino, line 80: you could use polly's map here

pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+	projectKeys := make(map[string]string, 0)
+	projects := make([]string, 0)
+	for _, p := range pr.Items {
+		projectKeys[p.Name] = p.Key
+		projects = append(projects, p.Name)
+	}
+
+	project := ""
+	if len(projects) == 0 {
+		return Context{}, fmt.Errorf("You don't have any Bitbucket project created. Please create one first \n")
```
michaeljguarino, line 113: reword to `You don't have a BitBucket project created, please create one and try again` perhaps

pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+		return Context{}, err
+	}
+	sshAddress, err := getSSHAddress(res.Links)
+	if err != nil {
+		return Context{}, err
+	}
+
+	utils.Highlight("Setting up a read-write deploy key for this repo...\n")
+
+	if _, err := b.Client.Repositories.DeployKeys.Create(&bitbucket.DeployKeyOptions{
```
michaeljguarino, line 156: do these have write permissions by default (will be needed)?

pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+	if err != nil {
+		return Context{}, err
+	}
+	defer os.RemoveAll(dir)
+	gitIgnore := path.Join(dir, ".gitignore")
+	if err := os.WriteFile(gitIgnore, []byte(""), 0644); err != nil {
+		return Context{}, err
+	}
+
+	if err := b.Client.Repositories.Repository.WriteFileBlob(&bitbucket.RepositoryBlobWriteOptions{
```
michaeljguarino, line 176: why do we need to do this?  `plural init` should override the gitignore (i remember another bitbucket user hitting this snag a looong time ago)

pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+
+func (b *Bitbucket) StarPluralGitHubRep() error {
+	return nil
+}
+
+func getSSHAddress(links map[string]interface{}) (string, error) {
+	sshAddress := ""
+	var linkList []Link
+	if links != nil {
+		clone := links["clone"]
```
michaeljguarino, line 205: what's going on here?  Seems very unusual we need to marshal then unmarshal whatever type this is

zreigz:
@michaeljguarino PTAL

pkg/scm/bitbucket.go
michaeljguarino: god, that is bonkers

pkg/scm/bitbucket.go
michaeljguarino: Is there no api that creates the repo with an initial commit?

pkg/scm/bitbucket.go
```diff
@@ -0,0 +1,249 @@
...
+		return Context{}, err
+	}
+	// The bitbucket creates empty repository.
+	// Initialize bitbucket repo with .gitignore file.
+	dir, err := os.MkdirTemp("", "bitbucket")
+	if err != nil {
+		return Context{}, err
+	}
+	defer os.RemoveAll(dir)
+	gitIgnore := path.Join(dir, ".gitignore")
```
michaeljguarino, line 171: Writing a basic README.md would be safer than a gitignore I think (which we do end up generating in `plural init`

pkg/scm/bitbucket.go
michaeljguarino: I feel like i've seen users have an initial commit, maybe it's set at a different level in bitbucket

zreigz: there is no API for this. Only UI has the option to create README and .gitignore files

pkg/scm/bitbucket.go
zreigz: done

PR #295, opened 2022-12-15
improve azure storage account creation

zreigz:
## Summary
Check if existing SA belongs to the resource group

## Checklist

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -25,6 +25,7 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
+	"github.com/azure/azure-sdk-for-go/services/compute/mgmt/2021-07-01/compute"
 	"github.com/pluralsh/plural/pkg/config"
 	"github.com/pluralsh/plural/pkg/kubernetes"
 	"github.com/pluralsh/plural/pkg/manifest"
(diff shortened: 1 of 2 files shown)


pkg/provider/azure.go
```diff
@@ -315,7 +319,27 @@ func (az *AzureProvider) Decommision(node *v1.Node) error {
...
 func (az *AzureProvider) getStorageAccount(account string) (storage.Account, error) {
-	return az.clients.Accounts.GetProperties(context.Background(), az.resourceGroup, account, storage.AccountExpandBlobRestoreStatus)
+	ctx := context.Background()
+	saList, err := az.clients.Accounts.List(ctx)
+	if err != nil {
+		return storage.Account{}, err
+	}
+	// check if storage account belongs to the same resource group
+	for _, sa := range saList.Values() {
+		if *sa.Name == account {
```
michaeljguarino, line 329: Why does this work?

zreigz: Dereferencing a pointer gives us access to the value the pointer points to

michaeljguarino: oh no not that part, ike how does the logic itself work?

michaeljguarino: You're listing storage accounts in the subscription then somehow getting the resource group by parsing the full path name or somnething?  Is there no reference to it elsewhere?

michaeljguarino: Wow, that's nuts. Azure really is trash then

michaeljguarino (approved):

PR #296, opened 2022-12-19
delete azure node

zreigz:
## Summary
The Azure API doesn't support the deletion of the node pool node. To do this we have to cordon, drain, then delete the node.

## Checklist

### DIFF
--- a/pkg/provider/azure.go
+++ b/pkg/provider/azure.go
@@ -20,7 +20,6 @@ import (
 	"github.com/Azure/go-autorest/autorest/azure"
 	"github.com/Azure/go-autorest/autorest/azure/auth"
 	"github.com/Azure/go-autorest/autorest/to"
-	"github.com/azure/azure-sdk-for-go/services/compute/mgmt/2021-07-01/compute"
 	"github.com/pluralsh/plural/pkg/config"
 	"github.com/pluralsh/plural/pkg/kubernetes"
 	"github.com/pluralsh/plural/pkg/manifest"
@@ -66,6 +65,7 @@ func GetClientSet(subscriptionId string) (*ClientSet, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	return &ClientSet{
 		Groups:         resourceGroupClient,
 		Accounts:       storageAccountsClient,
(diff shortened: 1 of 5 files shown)


pkg/provider/azure.go
```diff
@@ -303,15 +303,11 @@ func (azure *AzureProvider) Flush() error {
...
-	fut, err := vms.Delete(ctx, az.Project(), node.Name, to.BoolPtr(true))
+	kube, err := kubernetes.Kubernetes()
 	if err != nil {
-		return pluralerr.ErrorWrap(err, "failed to call deletion api")
+		return err
 	}
-
-	err = fut.WaitForCompletionRef(ctx, vms.Client)
-	return pluralerr.ErrorWrap(err, "vm deletion failed with")
+	return kube.DeleteNode(node)
```
michaeljguarino, line 310: Does this deprovision the vm?

zreigz: good point, l will check this

zreigz: @michaeljguarino unfortunately, the VMs are still there. I will check it again

michaeljguarino: yea i'm pretty sure we need to do what we were doing previously and kill the vm directly and allow azure's node groups to drain it automatically

zreigz: @michaeljguarino  Finally, I found a way to delete VM. We use `VirtualMachineScaleSets` for nodes. I was experimenting with the client and I found out that we need to deallocate the instance to scale down the set (otherwise will be recreated after deletion) and then we can delete the VM instance.

pkg/provider/azure.go
```diff
@@ -303,15 +308,61 @@ func (azure *AzureProvider) Flush() error {
...
+		InstanceIDs: []*string{
+			to.StringPtr(InstanceID)},
+	}, &armcompute.VirtualMachineScaleSetsClientBeginDeleteInstancesOptions{ForceDeletion: to.BoolPtr(true)})
+	if err != nil {
+		return err
+	}
+	_, err = pollerDelete.PollUntilDone(ctx, &runtime.PollUntilDoneOptions{
+		Frequency: 1 * time.Second,
+	})
+	if err != nil {
```
michaeljguarino, line 361: can't we just `return err` here instead of adding this branch?

michaeljguarino (approved):
Wow, this is a wildly unhelpful api

PR #301, opened 2023-01-03
ENG-1076 helm improvments

zreigz:
## Summary
Introduce `default-values.yaml` from the template to be able to override `values.yaml`
Migrate existing values file.

## Checklist

### DIFF
--- a/pkg/scaffold/helm.go
+++ b/pkg/scaffold/helm.go
@@ -88,6 +88,9 @@ func Notes(installation *api.Installation) error {
 	ctx, _ := context.Repo(installation.Repository.Name)
 	valuesFile := pathing.SanitizeFilepath(filepath.Join(repoRoot, repo, "helm", repo, "values.yaml"))
 	prevVals, _ := prevValues(valuesFile)
+	defaultValuesFile := pathing.SanitizeFilepath(filepath.Join(repoRoot, repo, "helm", repo, "default-values.yaml"))
+	defaultPrevVals, _ := prevValues(defaultValuesFile)
+
 	vals := map[string]interface{}{
 		"Values":        ctx,
 		"Configuration": context.Configuration,
@@ -117,6 +120,9 @@ func Notes(installation *api.Installation) error {
 		}
 	}
 
+	for k, v := range defaultPrevVals {
+		vals[k] = v
+	}
 	for k, v := range prevVals {
 		vals[k] = v
 	}
@@ -139,7 +145,16 @@ func Notes(installation *api.Installation) error {
 func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 	ctx, _ := w.Context.Repo(w.Installation.Repository.Name)
 	valuesFile := pathing.SanitizeFilepath(filepath.Join(s.Root, "values.yaml"))
+	defaultValuesFile := pathing.SanitizeFilepath(filepath.Join(s.Root, "default-values.yaml"))
+	defaultPrevVals, _ := prevValues(defaultValuesFile)
 	prevVals, _ := prevValues(valuesFile)
+
+	if !utils.Exists(valuesFile) {
+		if err := os.WriteFile(valuesFile, []byte(""), 0644); err != nil {
+			return err
+		}
+	}
+
 	conf := config.Read()
 
 	apps, err := NewApplications()
@@ -182,11 +197,14 @@ func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 		}
 	}
 
+	for k, v := range defaultPrevVals {
+		vals[k] = v
+	}
 	for k, v := range prevVals {
 		vals[k] = v
 	}
 
-	values, err := scftmpl.BuildValuesFromTemplate(vals, prevVals, w)
+	values, err := scftmpl.BuildValuesFromTemplate(vals, w)
 	if err != nil {
 		return err
 	}
@@ -198,7 +216,7 @@ func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 		return err
 	}
 
-	return utils.WriteFile(valuesFile, io)
+	return utils.WriteFile(defaultValuesFile, io)
 }
 
 func prevValues(filename string) (map[string]map[string]interface{}, error) {
--- a/pkg/scaffold/helm.go
+++ b/pkg/scaffold/helm.go
@@ -204,19 +204,49 @@ func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 		vals[k] = v
 	}
 
-	values, err := scftmpl.BuildValuesFromTemplate(vals, w)
+	defaultValues, err := scftmpl.BuildValuesFromTemplate(vals, w)
 	if err != nil {
 		return err
 	}
 
-	io, err := yaml.Marshal(values)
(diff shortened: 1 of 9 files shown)


pkg/scaffold/helm.go
```diff
@@ -139,7 +145,16 @@ func Notes(installation *api.Installation) error {
 func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 	ctx, _ := w.Context.Repo(w.Installation.Repository.Name)
 	valuesFile := pathing.SanitizeFilepath(filepath.Join(s.Root, "values.yaml"))
+	defaultValuesFile := pathing.SanitizeFilepath(filepath.Join(s.Root, "default-values.yaml"))
+	defaultPrevVals, _ := prevValues(defaultValuesFile)
 	prevVals, _ := prevValues(valuesFile)
+
+	if !utils.Exists(valuesFile) {
+		if err := os.WriteFile(valuesFile, []byte(""), 0644); err != nil {
```
DavidSpek, line 153: I'm not sure, but it's possible this might need to be `[]byte{"{}"}`.

pkg/wkspace/minimal.go
```diff
@@ -91,13 +93,16 @@ func (m *MinimalWorkspace) BounceHelm(extraArgs ...string) error {
 	if err == nil {
 		defer func(oldpath, newpath string) {
 			_ = os.Rename(oldpath, newpath)
-		}(backup, pathing.SanitizeFilepath(filepath.Join(path, "values.yaml")))
+		}(backup, pathing.SanitizeFilepath(filepath.Join(path, valuesYaml)))
```
DavidSpek, line 96: I'm pretty sure this should be the `default-values.yaml` file here. This is because at deployment time we have the ability to add values from terraform, but since those can change this is templated during the deploy stage. For example if the values file contains `example: '{{ .Import.Terraform.access_key_id }}'`. The workflow looks as follows:
- create a backup of the values file
- template the value from terraform into the values file so it conains `example: value`
- restore the backed up values file so that it contains `example: '{{ .Import.Terraform.access_key_id }}'` again and can be templated again during the next deployment (as the terraform output can change).

pkg/wkspace/minimal.go
```diff
@@ -53,7 +55,7 @@ func FormatValues(w io.Writer, vals string, output *output.Output) (err error) {
 
 func templateVals(app, path string) (backup string, err error) {
 	root, _ := utils.ProjectRoot()
-	valsFile := pathing.SanitizeFilepath(filepath.Join(path, "values.yaml"))
```
DavidSpek, line 56: This would then also need to be `default-values.yaml`.

pkg/wkspace/minimal.go
```diff
@@ -113,7 +118,7 @@ func (m *MinimalWorkspace) TemplateHelm() error {
 	if err == nil {
 		defer func(oldpath, newpath string) {
 			_ = os.Rename(oldpath, newpath)
-		}(backup, pathing.SanitizeFilepath(filepath.Join(path, "values.yaml")))
```
DavidSpek, line 116: This should also be `default-values.yaml`

pkg/wkspace/minimal.go
```diff
@@ -113,7 +118,7 @@ func (m *MinimalWorkspace) TemplateHelm() error {
...
+		}(backup, pathing.SanitizeFilepath(filepath.Join(path, valuesYaml)))
 	}
 
 	namespace := m.Config.Namespace(m.Name)
```
DavidSpek, line 124: The `helm template` command should also include the following:

```go
if utils.Exists(pathing.SanitizeFilepath(filepath.Join(path, defaultValuesYaml))) {
	defaultArgs = append(defaultArgs, []string{"-f", filepath.Join(path, defaultValuesYaml), "-f", filepath.Join(path, valuesYaml)}...)
}
```

pkg/wkspace/minimal.go
```diff
@@ -134,7 +139,7 @@ func (m *MinimalWorkspace) DiffHelm() error {
 	if err == nil {
 		defer func(oldpath, newpath string) {
 			_ = os.Rename(oldpath, newpath)
-		}(backup, pathing.SanitizeFilepath(filepath.Join(path, "values.yaml")))
```
DavidSpek, line 137: This should also be `default-values.yaml`.

pkg/wkspace/minimal.go
```diff
@@ -134,7 +139,7 @@ func (m *MinimalWorkspace) DiffHelm() error {
...
+		}(backup, pathing.SanitizeFilepath(filepath.Join(path, valuesYaml)))
 	}
 
 	namespace := m.Config.Namespace(m.Name)
```
DavidSpek, line 145: The helm diff should also include the following:

```go
if utils.Exists(pathing.SanitizeFilepath(filepath.Join(path, defaultValuesYaml))) {
	defaultArgs = append(defaultArgs, []string{"-f", filepath.Join(path, defaultValuesYaml), "-f", filepath.Join(path, valuesYaml)}...)
}
```

pkg/utils/map.go
```diff
@@ -14,6 +19,54 @@ func CleanUpInterfaceMap(in map[interface{}]interface{}) map[string]interface{}
...
+		}
+
+		t, ok := v.Interface().(map[string]interface{})
+		if ok {
+			RemoveNulls(t)
+		}
+	}
+}
+
+func PatchInterfaceMap(defaultValues, values map[string]map[string]interface{}) (map[string]map[string]interface{}, error) {
```
michaeljguarino, line 38: Why are we using this approach here?

zreigz: this is for migration from values.yaml to default-values.yaml. We make patch from those two files to find the only changes (value changes, new fields) and then we override values.yaml with those changes

zreigz: right now the `values.yaml` contains only changes to override

michaeljguarino: in that case, if this is just for migration purposes, we should have some conditional code to ensure we only do it when the default-values.yaml file has not yet been created.

michaeljguarino: Generally i think this should be programmed as defensively as possible

DavidSpek: The way it runs now actually will allow us to retroactively clean up the values files, which some apps can definitely benefit from. The migration also only happens for keys that have the same value that we template. So for example, if a user changes the OIDC client ID and secret in the `values.yaml` file that will not get migrated to the `default-values.yaml`.

michaeljguarino: Will we ever need to do this beyond the one initial time the new default-values.yaml file is created though?  If not, we should disable the code path from then on, by checking if it exists and ignoring.

DavidSpek: At the moment there are a number of apps we could do this for because we were lacking this functionality previously. One example is Argo-CD or Plural itself. For those applications we could push a chart version including the old templated values to get those into the `default-values.yaml` and then another update to clean them up.

DavidSpek: I've also checked it with arrays and if people customize an array that we template it is left in the `values.yaml` properly, overriding what we template as would be expected.

DavidSpek (approved):
Tested this a couple different ways and all is working as expected.

PR #306, opened 2023-01-11
fix: get helm values from adjacent apps + clean up empty nested maps

DavidSpek:
## Summary
After https://github.com/pluralsh/plural-cli/pull/301, the ability to get helm values from adjacent applications broke since it was still trying to get those from the `values.yaml` and not a combination of the `default-values.yaml` and `values.yaml`. This PR fixes this. Along the way, it occurred that empty nested maps could be left in the `values.yaml`, which has also been solved.

TODO: we still need to add unit tests for cleaning up the empty maps.

## Test Plan
- Manually test with the apps that were causing problems

## Checklist

### DIFF
--- a/pkg/scaffold/application.go
+++ b/pkg/scaffold/application.go
@@ -29,15 +29,14 @@ func NewApplications() (*Applications, error) {
 }
 
 func (apps *Applications) HelmValues(app string) (map[string]interface{}, error) {
-	res := make(map[string]interface{})
 	valuesFile := pathing.SanitizeFilepath(filepath.Join(apps.Root, app, "helm", app, "values.yaml"))
 	vals := make(map[string]interface{})
 	valsContent, err := os.ReadFile(valuesFile)
 	if err != nil {
-		return res, err
+		return nil, err
 	}
 	if err := yaml.Unmarshal(valsContent, &vals); err != nil {
-		return res, err
+		return nil, err
 	}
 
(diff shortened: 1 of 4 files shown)


pkg/scaffold/application.go
```diff
@@ -28,14 +29,35 @@ func NewApplications() (*Applications, error) {
...
+		if err := yaml.Unmarshal(defaultValsContent, &defaultVals); err != nil {
+			return res, err
+		}
+	}
 
-	err = yaml.Unmarshal(content, &res)
+	for k, v := range defaultVals {
+		res[k] = v
+	}
+	for k, v := range vals {
```
michaeljguarino, line 58: isn't this merge technically too strict?  Like it wouldn't handle nested objects properly

zreigz: it's not the typical merge. We have to override default values with values

DavidSpek: Yeah I wasn't happy about that merge, but got caught up removing empty maps properly. Lukasz change should have fixed this merging.

@zreigz Wouldn't what we do in https://github.com/pluralsh/plural-cli/blob/main/pkg/scaffold/helm.go#L200 also be too strict then?

zreigz (approved):
lgtm

pkg/scaffold/application.go
zreigz: @DavidSpek  here is another problem, I will make another PR for this

PR #307, opened 2023-01-12
don't merge previous values

zreigz:
## Summary
We should take previous values from `default-values.yaml` if it exists. Otherwise from `values.yaml` (old CLI or new deployment)
Before this fix, we could override some values. We don't need to merge previous values from  `default-values.yaml` and `values.yaml` because it's happening during helm execution `helm upgrade --install`

## Checklist


DavidSpek (approved):
This is probably the behavior that should be expected.

PR #310, opened 2023-01-12
Handle creds fetching for zonal vs regional gke clusters

michaeljguarino:
## Summary
You need to specify the zone to the cli for this, so use a dedicated context var to grab it

## Test Plan
regression tested a standard gcp workspace

## Checklist

### DIFF
--- a/pkg/provider/gcp.go
+++ b/pkg/provider/gcp.go
@@ -221,9 +221,10 @@ func (gcp *GCPProvider) KubeConfig() error {
 		return nil
 	}
 
+	t, name := gcp.clusterLocation()
 	cmd := exec.Command(
 		"gcloud", "container", "clusters", "get-credentials", gcp.Clust,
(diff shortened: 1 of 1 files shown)


pkg/provider/gcp.go
```diff
@@ -270,6 +271,14 @@ func (gcp *GCPProvider) mkBucket(name string) error {
 	return nil
 }
 
+func (gcp *GCPProvider) clusterLocation() (string, string) {
+	if z, ok := gcp.ctx["clusterZone"]; ok {
```
DavidSpek, line 275: Shouldn't we be setting this in the context somewhere?

michaeljguarino: it has to be done manually, and we'll document it as part of the process

DavidSpek: We should be able to handle that more gracefully but it’s fine for now.

DavidSpek (approved):
LGTM

PR #312, opened 2023-01-17
AWS BYOK

DavidSpek:
## Summary
This PR expands the AWS provider so that it allows for people to bring existing cluster and use them with Plural. If a user wants to use an existing cluster they will be able to select a cluster from the region they have choses. Along with this, the max length for cluster names has been expanded to what the providers actually support and this is truncated for the bucket names.

## Test Plan
Build CLI locally and test with new and existing clusters that are created by us as well as existing byok clusters.

## Checklist


pkg/provider/aws.go
```diff
@@ -60,34 +66,73 @@ var (
...
+
+		provider.ByokConf = &manifest.ByokConfig{Enabled: !createCluster, Type: EKS}
+
+		eksClient, err2 := getEksClient(provider.Reg, *provider.goContext)
+		if err != nil {
+			return nil, err2
+		}
+
+		input := &eks.ListClustersInput{}
+		clusters, err3 := eksClient.ListClusters(ctx, input)
```
michaeljguarino, line 115: we should factor this into a separate method probably to keep this one less sprawling

pkg/provider/aws.go
```diff
@@ -213,6 +271,16 @@ func (aws *AWSProvider) Context() map[string]interface{} {
 	return map[string]interface{}{}
 }
 
+func (aws *AWSProvider) Byok() map[string]interface{} {
```
michaeljguarino, line 274: is there a reason we can't just use the struct as the return val here?

pkg/utils/string.go
```diff
@@ -14,3 +14,13 @@ func Pluralize(one, many string, count int) string {
 func ToString(val interface{}) string {
 	return fmt.Sprintf("%v", val)
 }
+
+func TruncString(s string, c int) string {
```
michaeljguarino, line 18: it might be nice to move these to the polly lib at some point, but not crucial

pkg/provider/aws.go
```diff
@@ -60,34 +66,73 @@ var (
...
+
 func mkAWS(conf config.Config) (provider *AWSProvider, err error) {
 	provider = &AWSProvider{}
-	if err = survey.Ask(awsSurvey, provider); err != nil {
+	if err = survey.Ask(initAwsSurvey, provider); err != nil {
+		return
+	}
+
+	var createCluster bool
+	if err = survey.AskOne(&survey.Confirm{Message: "Do you want to create a new cluster?", Default: true}, &createCluster); err != nil {
```
michaeljguarino, line 92: We should also add to this note that if they don't create a cluster, they should only chose a newly created, fresh cluster

pkg/utils/string.go
DavidSpek: I was actually wanting something more advanced than a simple truncate but couldn't find something better.

pkg/provider/aws.go
DavidSpek: That is what I was trying to do first which also didn't need this function, but for some reason only the values were being returned during templating. But this is indeed something that I think should be cleaned up.

pkg/provider/aws.go
DavidSpek: We can add the note, but I'm not sure how strictly necessary this is since clashing applications can be disabled. I didn't add it in the artifact (yet), but I think the only one left is cert-manager. External DNS is limited to the subdomain set in the workspace so might not clash, but potentially could so we can add a way to disable that as well.

michaeljguarino: I'm thinking more if you have the zalando postgres operator or nginx ingress controller already installed

PR #322, opened 2023-01-31
[WIP] introduce terraform exec lib

zreigz:
## Summary
introduce terraform exec lib for terraform commands

## Checklist


cmd/plural/terraform.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/pluralsh/plural/pkg/terraform"
+	"github.com/pluralsh/plural/pkg/utils"
+	"github.com/urfave/cli"
+)
+
+func terraformCommands() []cli.Command {
+	return []cli.Command{
+		{
+			Name:   "init",
```
michaeljguarino, line 14: could we rename this to install?  Since it could be confounded with `terraform init` which does something a bit different

PR #329, opened 2023-02-02
fix terraform inputs templating

zreigz:
## Summary
fix terraform inputs templating

## Checklist

### DIFF
--- a/pkg/wkspace/minimal.go
+++ b/pkg/wkspace/minimal.go
@@ -1,6 +1,7 @@
 package wkspace
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/wkspace/minimal.go
```diff
@@ -120,8 +120,17 @@ func getValues(name string) (map[string]interface{}, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	backup, err := templateVals(name, path)
```
michaeljguarino, line 124: Shouldn't we bail from this if `err != nil`?

zreigz: we override the original values with inputs from terraform, later we have to revert this to the original one. We can do this when there are no errors. Generally, this is the legacy code I accidentally removed

michaeljguarino (approved):

PR #340, opened 2023-02-16
Fix silent errors on configuration failure

michaeljguarino:
## Summary
We would swallow config failures before instead of surfacing errors.  Also the NOT condition has a bug in its setup

## Test Plan
local + unit test

## Checklist


zreigz (approved):
lgtm

PR #342, opened 2023-02-20
add crypto key validation

zreigz:
## Summary
 - Create a `.keyid` file w/ the sha256 digest of the encryption key.
 - Fix the problem with the `workspace.yaml` created outside the git repository
-  Add new command to generate key fingerprint file: `plural crypto key-fingerprint`

## Checklist

### DIFF
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -144,7 +144,7 @@ func (p *Plural) cryptoCommands() []cli.Command {
 			Subcommands: p.backupCommands(),
 		},
 		{
-			Name:   "key-fingerprint",
+			Name:   "fingerprint",
 			Usage:  "generates a file with the key fingerprint",
 			Action: keyFingerprint,
 		},
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -180,13 +180,20 @@ func handleEncrypt(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-
-	prov, err := crypto.Build()
+	cryptoProv, err := crypto.Build()
+	if err != nil {
+		return err
+	}
+	keyID, err := crypto.GetKeyID()
 	if err != nil {
 		return err
 	}
 
-	result, err := crypto.Encrypt(prov, data)
+	if keyID != "" && cryptoProv.ID() != keyID {
+		return fmt.Errorf("the key fingerprint doesn't match")
+	}
+
+	result, err := crypto.Encrypt(cryptoProv, data)
 	if err != nil {
 		return err
 	}
@@ -321,6 +328,19 @@ func (p *Plural) handleSetupKeys(c *cli.Context) error {
 }
 
 func handleUnlock(c *cli.Context) error {
+	cryptoProv, err := crypto.Build()
+	if err != nil {
+		return err
+	}
+	keyID, err := crypto.GetKeyID()
+	if err != nil {
+		return err
+	}
+
+	if keyID != "" && cryptoProv.ID() != keyID {
+		return fmt.Errorf("the key fingerprint doesn't match")
+	}
+
 	repoRoot, err := git.Root()
 	if err != nil {
 		return err
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -143,6 +143,11 @@ func (p *Plural) cryptoCommands() []cli.Command {
 			Usage:       "manages backups of your encryption keys",
 			Subcommands: p.backupCommands(),
 		},
+		{
+			Name:   "key-fingerprint",
+			Usage:  "generates a file with the key fingerprint",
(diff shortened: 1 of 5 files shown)


pkg/provider/aws.go
```diff
@@ -276,3 +282,7 @@ func GetAwsAccount() (string, error) {
 	}
 	return res.Account, nil
 }
+
+func (aws *AWSProvider) KeyID() string {
```
michaeljguarino, line 286: I don't believe this should be set on the cloud provider actually (I mean we can but it should be orthogonal to it).  Is there a way we can remove the dependency there?

cmd/plural/crypto.go
```diff
@@ -180,13 +182,20 @@ func handleEncrypt(c *cli.Context) error {
...
+	if err != nil {
+		return err
+	}
+	prov, err := provider.GetProvider()
 	if err != nil {
 		return err
 	}
 
-	result, err := crypto.Encrypt(prov, data)
+	if prov.KeyID() != "" && cryptoProv.ID() != prov.KeyID() {
```
michaeljguarino, line 194: we should probably also do this validation on `plural crypto unlock` too

michaeljguarino: If there's a chicken-egg with the `workspace.yaml` being encrypted, then we should maybe create a new file in the repo for the key digest that is outside the encrypted list

zreigz: yes, good point

pkg/provider/aws.go
zreigz: we can remove it, especially when we would like to use a new file outside the encrypted list

zreigz:
@michaeljguarino PTAL

cmd/plural/init.go
```diff
@@ -43,9 +43,14 @@ func handleInit(c *cli.Context) error {
 	if err != nil {
 		return err
 	}
-	defer func(prov provider.Provider) {
+	kv, err := crypto.GenerateKeyValidator()
```
michaeljguarino, line 46: we should also run this in the setup phase of the cloud shell, and also have an independent crypto command to generate it manually for old repos (i'll maybe call that in the console or something)

michaeljguarino: @zreigz It's in `pkg/server/setup.go`

pkg/crypto/key.go
```diff
@@ -138,3 +138,56 @@ func (k *AESKey) Flush() error {
 
 	return os.WriteFile(getKeyPath(), io, 0644)
 }
+
+func (k *AESKey) ID() string {
+	sha := sha256.Sum256([]byte(k.Key))
```
michaeljguarino, line 143: do we have any utils code for sha'ing currently?  I feel like it's around somewhere

michaeljguarino:
Another thing I realized, and I think you probably covered but we should double-check, you should make sure the sha format here matches what's done in `plural crypto backups list` (you can verify by just creating a backup first w/ `plural crypto backups create`)

michaeljguarino:
So I was able to test it but I realized there's a lot of incompatibilities introduced w/ the new chart, it might be better if we take on the upgrade since it'll be more advanced. It can probably be done in a few days.

zreigz:
> 
Unfortunately, the SHA format is different than for `plural crypto backups` but I am not sure if it's relevant. We compare them in the flight using the same generation function

pkg/crypto/key.go
zreigz: no, we don't have

cmd/plural/crypto.go
```diff
@@ -143,6 +143,11 @@ func (p *Plural) cryptoCommands() []cli.Command {
 			Usage:       "manages backups of your encryption keys",
 			Subcommands: p.backupCommands(),
 		},
+		{
+			Name:   "key-fingerprint",
```
michaeljguarino, line 147: Perhaps just name this `fingerprint` to be a bit more terse?

cmd/plural/init.go
```diff
@@ -43,20 +43,22 @@ func handleInit(c *cli.Context) error {
...
 		repo, err = scm.Setup()
 		if err != nil {
 			return err
 		}
 		gitCreated = true
-	} else if !git && err != nil {
-		return err
+	}
+	if !git && !gitCreated {
+		return fmt.Errorf("you're not in the git repository")
```
michaeljguarino, line 55:
```
you're not in a git repository, either clone one directly or let us set it up for you by rerunning `plural init`
````

michaeljguarino (approved):
small language nits, but otherwise this is good to go

michaeljguarino:
@zreigz so the difference between the server and here is that we base32 encode it instead of base64 encode it.  If you swap the schemes I think that will keep them aligned

zreigz:
@michaeljguarino I have changed to base32 and they match, thanks

PR #346, opened 2023-02-22
fix plural diff

zreigz:
## Summary
Fix getting the repo path for some helm commands

## Checklist

### DIFF
--- a/pkg/wkspace/minimal.go
+++ b/pkg/wkspace/minimal.go
@@ -121,7 +121,7 @@ func getValues(name string) (map[string]interface{}, error) {
 	values := make(map[string]interface{})
 	defaultVals := make(map[string]interface{})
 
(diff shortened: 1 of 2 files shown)


pkg/wkspace/minimal.go
```diff
@@ -295,3 +295,21 @@ func templateTerraformInputs(name, vals string) ([]byte, error) {
 	}
 	return buf.Bytes(), nil
 }
+
+func getRepoPath(name string) (string, error) {
+	wd, err := os.Getwd()
```
michaeljguarino, line 300: Would it be better to use the ProjectRoot helper to just find the appropriate root then generate the path from there?   This logic seems brittle

zreigz: thanks, done

DavidSpek (approved):
LGTM

PR #350, opened 2023-02-23
Check for using root user

michaeljguarino:
## Summary
Apparently a user tried this and aws doesn't support policy simulation, we should just explain that's not what you should be doing period

## Test Plan
local

## Checklist


zreigz (approved):
lgtm

PR #351, opened 2023-02-23
Generate useful repo README file

michaeljguarino:
## Summary

This could potentially help onboard users quicker to understanding their repo. Refactored to just pull from our docs github so we can easily iterate on it as well.

## Test Plan
executed `plural readme` to confirm it generates appropriately

## Checklist

### DIFF
--- a/pkg/server/setup.go
+++ b/pkg/server/setup.go
@@ -13,6 +13,7 @@ import (
 	"github.com/pluralsh/plural/pkg/manifest"
 	"github.com/pluralsh/plural/pkg/provider"
 	"github.com/pluralsh/plural/pkg/utils"
+	"github.com/pluralsh/plural/pkg/wkspace"
 )
 
 func toConfig(setup *SetupRequest) *config.Config {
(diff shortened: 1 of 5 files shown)


DavidSpek:
What happens if the file already exists?

michaeljguarino:
It would just overwrite it currently, but the only paths it executes currently is init and shell setup, so really shouldn't overwrite ever

pkg/server/setup.go
```diff
@@ -118,6 +119,7 @@ func setupCli(c *gin.Context) error {
 	if err := crypto.CreateKeyFingerprintFile(); err != nil {
 		return err
 	}
+	_ = wkspace.DownloadReadme()
```
zreigz, line 122: why did you hide the error?

DavidSpek (approved):
LGTM

swoodward90 (approved):

pkg/server/setup.go
michaeljguarino: it's on a critical path and I don't want init to fail if it fails (eg the user still sets up but maybe doesn't have a readme in their repo, no big deal)

PR #352, opened 2023-02-24
Auto-assume service account in installation repo

zreigz:
## Summary
If the owner's email of the project is not the current user, impersonate that user and update go client.

## Checklist

### DIFF
--- a/cmd/plural/validation.go
+++ b/cmd/plural/validation.go
@@ -107,13 +107,13 @@ func tracked(fn func(*cli.Context) error, event string) func(*cli.Context) error
 func validateOwner() error {
 	path := manifest.ProjectManifestPath()
 	project, err := manifest.ReadProject(path)
-	conf := config.Read()
 	if err != nil {
 		return fmt.Errorf("Your workspace hasn't been configured. Try running `plural init`.")
 	}
 
 	if owner := project.Owner; owner != nil {
-		if owner.Email != conf.Email || owner.Endpoint != conf.Endpoint {
+		conf := config.Read()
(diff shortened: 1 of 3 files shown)


cmd/plural/validation.go
```diff
@@ -192,3 +161,36 @@ func upstreamSynced(fn func(*cli.Context) error) func(*cli.Context) error {
...
 	}
 }
+
+func (p *Plural) impersonate(fn func(*cli.Context) error) func(*cli.Context) error {
+	return func(c *cli.Context) error {
+		p.InitPluralClient()
+		path := manifest.ProjectManifestPath()
+		project, err := manifest.ReadProject(path)
+		conf := config.Read()
+		if err != nil {
```
michaeljguarino, line 171: We probably shouldn't error here, or is it only used when the repo is already set up?

michaeljguarino (commented):
One thing we need to add here is there are other fetches to the config w/in the code outside client initialization.  The biggest I can think of is scaffolding terraform and helm but we should double check other possibilities too

michaeljguarino:
I think if when we assume a service account, a global variable within the `pkg/config` package is set and any future `config.Read()` function returns it if non-null it might guarantee all code paths are auto-wired with the service account.  Or at least that's how i guessed an elegant solve might work

cmd/plural/validation.go
zreigz: yes, it's used when the repo is already set up

michaeljguarino (approved):
lgtm

PR #353, opened 2023-02-27
Fail 'plural init' if a user is currently running a gcp demo

zreigz:
## Summary
Fail 'plural init' if a user is currently running a GCP demo

## Checklist


cmd/plural/init.go
```diff
@@ -21,10 +21,21 @@ import (
...
 
-func handleInit(c *cli.Context) error {
+const DemoingErrorMsg = "You're currently running a gcp demo cluster. Spin that down at https://app.plural.sh/shell before beginning a local installation"
+
+func (p *Plural) handleInit(c *cli.Context) error {
+	p.InitPluralClient()
 	gitCreated := false
 	repo := ""
 
+	me, err := p.Me()
```
michaeljguarino, line 31: I'm pretty sure this needs to be done after `handleLogin` (a fresh user won't have a way to query the api before)

michaeljguarino (changes requested):

cmd/plural/init.go
zreigz: yes, you are right

zreigz:
@michaeljguarino I have applied your comment. I will add unit and e2e tests later because it requires some refactoring.

cmd/plural/init.go
```diff
@@ -21,7 +21,10 @@ import (
 	"github.com/pluralsh/plural/pkg/wkspace"
 )
 
-func handleInit(c *cli.Context) error {
+const DemoingErrorMsg = "You're currently running a gcp demo cluster. Spin that down at https://app.plural.sh/shell before beginning a local installation"
+
+func (p *Plural) handleInit(c *cli.Context) error {
+	p.InitPluralClient()
```
michaeljguarino, line 27: This doesn't have any issues when there's no config created yet?

zreigz: done

michaeljguarino (approved):

PR #354, opened 2023-02-27
fix sa auto-assume

michaeljguarino:
## Summary
the previous implementation would auto-assume for any command whatsoever, even for profile commands for instance which would be used to switch back to your primary login

## Test Plan
local

## Checklist

- [x] I have added tests to cover my changes.


zreigz (approved):
/lgtm

PR #367, opened 2023-03-10
introduce posthog

zreigz:
## Summary
Introduce posthog for plural bundle install, build, and deploy commands. 
```
client.Enqueue(posthog.Capture{
  DistinctId: "someID",
  Event:      "cli_deploy",
  Properties: map[string]interface{}{
        "clusterName":        "test-cluster",
        "applicationName":    "bootstrap",
        "applicationID":      "123456-abcdef"
        "packageType":         "Helm",
        "packageName":         "bootstrap",
        "packageId":           "123456-abcdef",
        "packageVersion":      "0.1.0",
        "provider":           "AWS",
        "error":              "some error about the helm install failing",
    }
})
```

## Checklist


cmd/plural/deploy.go
```diff
@@ -247,7 +250,25 @@ func (p *Plural) deploy(c *cli.Context) error {
 				fmt.Println("")
 			}
 		}
-
+		for _, ch := range man.Charts {
```
michaeljguarino, line 253: We should send this info from the server if we want it in posthog, having the cli send it is error prone and could potentially cause failed installs

pkg/scaffold/scaffold.go
```diff
@@ -120,6 +121,17 @@ func (s *Scaffold) Execute(wk *wkspace.Workspace, force bool) error {
 			preflight.Sha = ""
 		}
 
+		for _, ch := range wk.Charts {
```
michaeljguarino, line 124: Similarly, this is better to send server-side

pkg/bundle/installer.go
```diff
@@ -73,13 +73,22 @@ func doInstall(client api.Client, recipe *api.Recipe, repo, name string, refresh
...
+		RecipeName:      recipe.Name,
+	}
 	if err := performTests(context, recipe); err != nil {
+		posthogProperty.Error = fmt.Errorf("failed to perform tests")
+		utils.PosthogCapture(utils.InstallPosthogEvent, posthogProperty)
 		return err
 	}
 
 	if err := client.InstallRecipe(recipe.Id); err != nil {
+		posthogProperty.Error = fmt.Errorf("failed install recipe")
```
michaeljguarino, line 87: should we just add this to our existing error logging wrapper?

cmd/plural/deploy.go
DavidSpek: We can't track the actual commands users execute server side since it doesn't know when somebody runs `plural deploy`. Having it in the CLI allows us to have a better metric of which commands people run, how often they run, and if they encountered an error when running the command. This can give us a metric of how many successful deploys have happened versus how many errors have occurred. It can also give us a timeline of a user's actions. For example, user deploys but runs into an error for app X, user then runs `install` to change a config and `build` for app X to fix that config, user then runs `deploy` again and it runs successfully.

Each command that a user executes and if it passes or fails is very important data so we can have a better idea of how many command executions pass versus fail and where they fail exactly.

pkg/bundle/installer.go
zreigz: done

zreigz: I have moved error handling to `tracked` validator

PR #370, opened 2023-03-15
add validator to update kubeconfig

zreigz:
## Summary
Update kubeconfig before all kube commands, crd and helm install

## Checklist


michaeljguarino (commented):
I think you need to cover `plural watch` too

michaeljguarino (commented):
I think this would also be needed in the plain `plural bounce` command too

zreigz:
@michaeljguarino PTAL

michaeljguarino (approved):
We should follow onto this and see if there's a way to not actually run the aws/gcloud/etc command if its been run before, and just do a kubeconfig context switch, but that can be done later

PR #371, opened 2023-03-15
feat: Add embedded UI with app installer wizard

floreks:
## Summary

- Reorganized project structure and exported `main.go` to root. Unfortunately, wails framework enforces project layout as it doesn't expose any configuration to allow changing entrypoint.
- Updated CD pipeline
  - Used `goreleaser-pro` split build feature to configure parallel multi-OS build. It's best to build on native OS when wails UI is being embedded into the binary. 
  - Updated GitHub action to support new configuration
- Updated local makefile targets to simplify development
  - `make build-cli` - builds CLI binary for the host OS without embedded UI
  - `make build-cli-ui` - builds CLI binary for the host OS with embedded UI
  - `make run-web` - runs dev env for the UI
  - `make generate-bindings` - generates frontend bindings for public Go methods provided as part of the app configuration. They are generated in `pkg/ui/web/wailsjs` directory
- Updated Dockerfiles to not include UI into the binary during build and build from root dir
- Used a conditional build to only embed the UI when `ui` tag is provided for the go build
- Bootstraped the UI react project similar to our other projects. It's in `pkg/ui/web` directory
- Fixed an issue with CLI docker container not working due to missing permissions to run the CLI binary
- UI offers an application installer wizard similar to both plural app and console
- Added a new subcommand `plural install` that opens an embedded UI

### Architecture support
In general embedded UI will only be available for below architectures:
- `linux/amd64`
- `windows/amd64`
- `darwin/amd64`
- `darwin/arm64`

Other supported architectures (`linux/arm64` and `windows/arm64`) and docker images will not have embedded UI.

## Test Plan

Locally and partially via GitHub action that stored output binaries for different architectures. I was able to test them on both windows/linux and with help of @maciaszczykm on M1 macOS.

## Checklist


pkg/ui/client.go
```diff
@@ -0,0 +1,187 @@
...
+	if err != nil {
+		return api.GetErrorResponse(err, "GetRecipeByID")
+	}
+
+	// Merge incoming configuration with existing one and update context
+	maps.Copy(mergedConfiguration, configuration)
+	context.Configuration[repoName] = mergedConfiguration
+
+	// Configure OIDC if enabled
+	if oidc {
```
michaeljguarino, line 141: this should happen after installing the recipe i believe

floreks: ahh, I thought it doesn't do any API mutations so I moved it there. I see it does the `UpsertOidcProvider` after all. I will change it then.

michaeljguarino (approved):
Let's merge it after fixing the issues w/ the go.mod and go.sum files

PR #397, opened 2023-04-13
fix: Add preflight checks and improve error handling in the UI

floreks:
## Summary

- Added some preflight checks before running `plural install` UI:
  - check if inside a git repo (`wkspace.Preflight`)
  - check if `context.yaml` exists
  - check if `workspace.yaml` exists
- Added global apollo client error handler to display toast with error message.
- Fixed an issue with provider name not being normalized for further GQL API usage (this broke UI on GCP provider)
- Fixed an issue with grommet fileinput not working properly in production build (worked in dev)
- Fixed an issue with configure OIDC not respecting confirm option

## Test Plan

Locally

## Checklist


pkg/bundle/oidc.go
```diff
@@ -116,6 +116,10 @@ func formatRedirectUris(settings *api.OIDCSettings, ctx map[string]interface{})
 }
 
 func confirmOidc(confirm *bool) (bool, error) {
+	if confirm != nil && *confirm {
```
floreks, line 119: It didn't really make sense to keep it at the end of function since it was never really used.

michaeljguarino:
I actually think you just need to wrap the function in the `rooted` validator like here: https://github.com/pluralsh/plural-cli/blob/main/cmd/plural/plural.go#L91 and it'll provide our standard error message when the command is run outside the repo

floreks:
Done. I have also wrapped it with `tracked` method.

michaeljguarino (approved):

PR #401, opened 2023-05-02
Minor bug fix: don't write empty encrypted files, they confuse Git

maciejp-ro:
## Summary

I bumped into a quirk with Git encryption filter: for some reason, `git diff --stat` and family doesn't call `textconv` if the decrypted file in repo is empty. It compares the decrypted empty file with encrypted file in repo, which results in a permanent phantom diff.

Which would be a minor quirk, but `git diff --shortstat` is what [my zsh prompt](https://github.com/nojhan/liquidprompt) calls to get summary, which means my shell thinks there are some changes and my prompt is red, which is personally annoying.

Liquidprompt is quite popular and it's likely that other similar status tools for shells, editors and such are confused.

All the empty files (so far) were `values.yaml` in Helm directories. This PR lets `yaml.Marshal` write `{}` into the values file, which makes the files not empty and fixes the diff.

Drive-by fix: one unrelated test was failing for me because of a symlink in `$TMPDIR` (AFAIK a normal thing in MacOS), fixed that too.

## Labels
bug-fix

## Test Plan
Ran `plural build` in my repository, ran `make test`

## Checklist
- [x] I have added tests to cover my changes.

### DIFF
--- a/pkg/scaffold/helm.go
+++ b/pkg/scaffold/helm.go
@@ -149,7 +149,7 @@ func (s *Scaffold) buildChartValues(w *wkspace.Workspace) error {
 	prevVals, _ := prevValues(valuesFile)
 
 	if !utils.Exists(valuesFile) {
-		if err := os.WriteFile(valuesFile, []byte(""), 0644); err != nil {
+		if err := os.WriteFile(valuesFile, []byte("{}\n"), 0644); err != nil {
 			return err
 		}
 	}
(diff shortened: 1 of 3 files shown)


maciejp-ro:
(I cannot add actual labels to the PR, but added them in the body as suggested by the PR template)

michaeljguarino (approved):

michaeljguarino:
Thanks for taking this on, we'll get it out and released asap

PR #410, opened 2023-05-31
fix: Fix archive creation during release

floreks:
## Summary

This should fix our CD pipeline. Archive section needed some tuning to support builds array.

## Test Plan

Locally with `GITLAB_CLIENT_SECRET=x goreleaser release --snapshot --clean --split`

## Checklist

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -46,9 +46,7 @@ builds:
           - CGO_ENABLED=0
   # Build CLI binary without embedded UI for linux. Required by our Console.
   - id: plural-cli-console
-    targets:
-      - linux_amd64
-      - linux_arm64
+    targets: [linux_amd64]
     ldflags:
       - -s
       - -w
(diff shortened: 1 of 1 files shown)


.goreleaser.yaml
```diff
@@ -46,9 +46,7 @@ builds:
           - CGO_ENABLED=0
   # Build CLI binary without embedded UI for linux. Required by our Console.
   - id: plural-cli-console
-    targets:
-      - linux_amd64
-      - linux_arm64
+    targets: [linux_amd64]
```
DavidSpek, line 49: Why was `linux_arm64` removed here?

floreks: I have noticed that it already builds `linux_arm64` without the UI: https://github.com/pluralsh/plural-cli/pull/410/files#diff-7326b55c062b0f46fe9e39aace0a25f4515cf206040fb91a6fd2cae839f5e826R35-R41

No need to build the same binary twice.

michaeljguarino (approved):

PR #414, opened 2023-06-02
fix sharing plural repo

zreigz:
## Summary
 - fix sharing a plural repo
 - fix the `Invalid cross-device link` error for `os.Rename` function

## Checklist


cmd/plural/crypto.go
```diff
@@ -53,6 +53,7 @@ const Gitignore = `/**/.terraform
 *.swo
 .DS_STORE
 .vscode
+.keyid
```
michaeljguarino, line 56: we can't gitignore this, we want to have the fingerprint validation in place

PR #415, opened 2023-06-02
Fix order in which fingerprints are evaluated

michaeljguarino:
## Summary
this needs to be done on the key generated from the provider, not at the start

## Test Plan
tbd

## Checklist

### DIFF
--- a/cmd/plural/crypto.go
+++ b/cmd/plural/crypto.go
@@ -334,12 +334,16 @@ func handleUnlock(c *cli.Context) error {
 		return err
 	}
 
-	gitIndex, _ := filepath.Abs(filepath.Join(repoRoot, ".git", "index"))
-	dump, err := os.CreateTemp("", "index.bak")
+	// fixes Invalid cross-device link when using os.Rename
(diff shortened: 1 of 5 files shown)


zreigz:
@michaeljguarino This solution also doesn't work. We still compare two different encryption keys. The SHA is generated from `~/.plural/key` When we import some repo from another user the key is always different than ours. The validation was added to check if the current encryption key is different from the one that was used during repo init. The sharing scenario will always fail in this case. In my opinion, we should add a user email address to the `.keyid` file and validate the key when the emails are equal

michaeljguarino:
This should extract the key from the crypto provider, in the case of sharing that will be the key decrypted from w/in `.plural-crypt`.  Have you tested it to confirm it doesn't work?  If not we probably just need to tweak some of the validation

michaeljguarino:
@zreigz fixed the bug with latest commit

zreigz:
@michaeljguarino I have fixed a few issues and tested it again. Adde e2e test for repo sharing passed. PTAL and we can merge it

zreigz (approved):

PR #421, opened 2023-07-04
cleanup finalizers for all namespace resources

zreigz:
## Summary
Don't force destroy namespaces during `plural destroy`.
The new implementation search for all namespaced resource types and then remove finalizers for the remaining objects in the namespace.

## Checklist


michaeljguarino (commented):
How can we confirm this is effective?  I do want us to be sure we can do full destroys without issue as consistently as possible, since not being able to tear down a cluster you no longer want is no bueno

PR #422, opened 2023-07-12
feat: Populate app versions in app crds

michaeljguarino:
## Summary
We previously just used helm chart versions, but can dive into the Chart.yaml and grab app version where present

## Test Plan
verified w/ `plural build` locally

## Checklist

### DIFF
--- a/pkg/api/charts.go
+++ b/pkg/api/charts.go
@@ -138,6 +138,7 @@ func convertVersion(version *gqlclient.VersionFragment) *Version {
 	v := &Version{
 		Id:      version.ID,
 		Version: version.Version,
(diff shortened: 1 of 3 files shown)


DavidSpek:
If an app has multiple helm charts, how is the app version of the wrapper chart determined?

michaeljguarino:
> If an app has multiple helm charts, how is the app version of the wrapper chart determined?

There's an `application` attribute on the helm chart that dictates it and should be set uniquely by the packager.  Otherwise the logic just takes the first and it can be incorrect but won't fail at least.

zreigz (approved):
lgtm

PR #437, opened 2023-08-25
Modify aws auth configmap manually to solve migration chicken-egg

michaeljguarino:
## Summary
This allows us to reusably modify the aws-auth configmap for eks from the client which should help resolve some migration-time issues

## Test Plan
tested locally from subcommands

## Checklist

### DIFF
--- a/cmd/plural/clusters.go
+++ b/cmd/plural/clusters.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/pluralsh/plural/pkg/api"
 	"github.com/pluralsh/plural/pkg/bootstrap"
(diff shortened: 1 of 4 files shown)


DavidSpek:
I think it'd probably be better to handle this automatically in the migrator code. See: https://github.com/pluralsh/cluster-api-migration/blob/main/pkg/aws/migrator.go

michaeljguarino:
It looks like that's been built to mostly scrape cloud apis and generate a cluster spec, not modify k8s resources actually.  Also I do like having the top level aws-auth commands for future use.

DavidSpek (approved):
LGTM

PR #455, opened 2023-09-26
[WIP]create deployments commands

zreigz:
## Summary
New CLI commands for CD deployments
Set env variables:
PLURAL_CONSOLE_URL=https://console.cd-demo.onplural.sh
PLURAL_CONSOLE_TOKEN=xxxx

## Checklist


pkg/console/console.go
```diff
@@ -0,0 +1,30 @@
...
+	"context"
+	"fmt"
+	"net/http"
+
+	consoleclient "github.com/pluralsh/console-client-go"
+)
+
+type consoleClient struct {
+	ctx          context.Context
+	pluralClient *consoleclient.Client
```
michaeljguarino, line 13: should this just be named `client` or something not specifically plural

pkg/console/models.go
```diff
@@ -0,0 +1,89 @@
+package console
```
michaeljguarino, line 1: Would it be possible to avoid using a new layer of types and just use the console client types entirely?

pkg/console/agent.go
```diff
@@ -0,0 +1,138 @@
...
+	"helm.sh/helm/v3/pkg/storage/driver"
+	"sigs.k8s.io/yaml"
+)
+
+var settings = cli.New()
+
+const (
+	repoName  = "deployop"
+	repoUrl   = "https://pluralsh.github.io/deployment-operator"
+	namespace = "console"
```
michaeljguarino, line 30: the namespace for this should actually be `plrl-deploy-operator` (it'll be installed in non-plural clusters so that seemed more general)

michaeljguarino: we also are probably going to need to use the `create-namespace` flag as a result of that

pkg/console/services.go
```diff
@@ -56,13 +57,30 @@ func (c *consoleClient) CreateClusterService(clusterId, clusterName *string, att
...
 }
 
-func (c *consoleClient) UpdateClusterService(serviceId string, attributes gqlclient.ServiceUpdateAttributes) (*gqlclient.UpdateServiceDeployment, error) {
-	result, err := c.client.UpdateServiceDeployment(c.ctx, serviceId, attributes)
+func (c *consoleClient) UpdateClusterService(serviceId, serviceName, clusterName *string, attributes gqlclient.ServiceUpdateAttributes) (*gqlclient.ServiceDeploymentFragment, error) {
+	if serviceId == nil && serviceName == nil && clusterName == nil {
+		return nil, fmt.Errorf("serviceId, serviceName and clusterName can not be null")
+	}
+	if serviceId != nil {
+		result, err := c.client.UpdateServiceDeployment(c.ctx, *serviceId, attributes)
```
michaeljguarino, line 65: we should be able to merge these mutations i believe

michaeljguarino (approved):

PR #471, opened 2023-11-13
feat: cd install check cluster

zreigz:
## Summary
Check if the url and token are valid for the cluster

## Checklist

### DIFF
--- a/cmd/plural/cd.go
+++ b/cmd/plural/cd.go
@@ -62,9 +62,9 @@ func (p *Plural) cdCommands() []cli.Command {
 			},
 		},
 		{
-			Name:   "eject",
-			Action: p.handleEject,
-			Usage:  "ejects cluster scaffolds",
+			Name:      "eject",
+			Action:    p.handleEject,
+			Usage:     "ejects cluster scaffolds",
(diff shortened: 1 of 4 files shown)


cmd/plural/cd.go
```diff
@@ -93,6 +93,17 @@ func (p *Plural) doInstallOperator(url, token string) error {
...
 	}
+	consoleClient, err := console.NewConsoleClient(token, url)
+	if err != nil {
+		return err
+	}
+	myCluster, err := consoleClient.MyCluster()
+	if err != nil {
+		return err
+	}
+	if !confirm(fmt.Sprintf("Are you sure you want to install deploy operator for %s cluster?", myCluster.MyCluster.Name), "PLURAL_INSTALL_AGENT_CONFIRM") {
```
michaeljguarino, line 104: Could we perhaps print a more thorough description of the cluster, eg something close to its describe:

```
Name: <name>
Handle: <handle>
Provider: (if it exists)
```

That way users have a really transparent view of what the cluster is.

zreigz: image
done

michaeljguarino (approved):

PR #504, opened 2024-04-17
set default values for agent installation

zreigz:
## Summary
Use global helm values for agent during `plural cd clusters bootstrap` command

## Checklist

### DIFF
--- a/cmd/plural/cd.go
+++ b/cmd/plural/cd.go
@@ -4,13 +4,13 @@ import (
 	"fmt"
 	"os"
 
-	"github.com/urfave/cli"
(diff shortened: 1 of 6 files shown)


cmd/plural/cd.go
```diff
@@ -118,7 +118,17 @@ func (p *Plural) doInstallOperator(url, token, values string) error {
 	}
 
 	vals := map[string]interface{}{}
-	if values != "" {
+	if values == "" {
```
michaeljguarino, line 121: We actually probably would want to merge the cli flag provided values and the global values if both are provided

zreigz: done

michaeljguarino (approved):

PR #511, opened 2024-05-17
check if deployment operator already exists

zreigz:
## Summary
check if deployment operator already exists and confirm if so

## Checklist

### DIFF
--- a/cmd/plural/cd.go
+++ b/cmd/plural/cd.go
@@ -117,6 +117,15 @@ func (p *Plural) doInstallOperator(url, token, values string) error {
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


michaeljguarino (approved):

cmd/plural/cd.go
```diff
@@ -117,6 +117,15 @@ func (p *Plural) doInstallOperator(url, token, values string) error {
 	if err != nil {
 		return err
 	}
+	alreadyExists, err := console.IsAlreadyAgentInstalled(p.Kube.GetClient())
```
michaeljguarino, line 120: we potentially should add an argument here to not do this check on `plural cd clusters reinstall` (the expectation is you are updating an already existing agent)

PR #528, opened 2024-07-30
create PR automation

zreigz:
``` 
plural cr pr-autoamtion create ID --context
```

## Checklist


zreigz:
@michaeljguarino PTAL

cmd/plural/cd_prautomation.go
```diff
@@ -0,0 +1,63 @@
...
+	"os"
+
+	"github.com/pluralsh/plural-cli/pkg/utils"
+	"github.com/samber/lo"
+	"github.com/urfave/cli"
+)
+
+func (p *Plural) cdPrAutomations() cli.Command {
+	return cli.Command{
+		Name:        "pr-automation",
```
michaeljguarino, line 14: I think this should be within the `pr` subcommand (I think there's already one).  That way `plural pr create ...` reads a lot more clearly

zreigz: done

zreigz:
@michaeljguarino I had to migrate it to the new structure, PTAL

michaeljguarino (approved):
we should also have a way to create by automation name too (don't know if this would require fetching the pr automation first to grab its id though)

zreigz:
> we should also have a way to create by automation name too (don't know if this would require fetching the pr automation first to grab its id though)

I will create an issue for it

PR #531, opened 2024-08-19
add new binary pluralctl

zreigz:
## Summary
A new distribution of the CLI for fleet-management.

## Checklist

### DIFF
--- a/cmd/pluralctl/main.go
+++ b/cmd/pluralctl/main.go
@@ -1,23 +1,26 @@
 package main
 
 import (
+	"log"
+	"os"
+
 	"github.com/fatih/color"
 	"github.com/pluralsh/plural-cli/cmd/command/cd"
+	"github.com/pluralsh/plural-cli/cmd/command/clone"
 	cryptocmd "github.com/pluralsh/plural-cli/cmd/command/crypto"
+	"github.com/pluralsh/plural-cli/cmd/command/down"
 	cmdinit "github.com/pluralsh/plural-cli/cmd/command/init"
 	"github.com/pluralsh/plural-cli/cmd/command/mgmt"
 	"github.com/pluralsh/plural-cli/cmd/command/pr"
-	"github.com/pluralsh/plural-cli/cmd/command/stack"
+	"github.com/pluralsh/plural-cli/cmd/command/profile"
 	"github.com/pluralsh/plural-cli/cmd/command/up"
+	"github.com/pluralsh/plural-cli/cmd/command/version"
 	"github.com/pluralsh/plural-cli/pkg/client"
-	"github.com/pluralsh/plural-cli/pkg/common"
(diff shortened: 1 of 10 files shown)


cmd/pluralctl/main.go
```diff
@@ -0,0 +1,112 @@
...
+	app.Name = ApplicationName
+	app.Usage = "Tooling to manage and operate a fleet of clusters"
+	app.EnableBashCompletion = true
+	app.Flags = globalFlags()
+	commands := []cli.Command{
+		cryptocmd.Command(plural.Plural),
+		cd.Command(plural.Plural, plural.HelmConfiguration),
+		up.Command(plural.Plural),
+		pr.Command(),
+		stack.Command(plural.Plural),
```
michaeljguarino, line 99: Are these stacks from console or plural?  We'd only care about the console stack commands here, and I do think they deserve a separate subcommand.

cmd/pluralctl/main.go
```diff
@@ -0,0 +1,112 @@
...
+		cmdinit.Command(plural.Plural),
+		mgmt.Command(plural.Plural),
```
michaeljguarino, line 101: We should probably also keep:

* `profile` commands
* `clone` command (to help clone + decrypt repos)
* `version` command (always useful to have)

cmd/pluralctl/main.go
zreigz: those are plural stacks. We have console stacks in `cd` subcommand

michaeljguarino: Yeah in that case transfer out the ones from `cd` into the top-level here if possible.

michaeljguarino: the idea is the cli should sort of mimic the sidebar tab structure in the console ui.

zreigz: I will try to do this

zreigz:
@michaeljguarino 
I've moved CD commands to the top level.
I had to move the plural login to the management group commands (duplicated with the CD login)

cmd/pluralctl/main.go
michaeljguarino: We do still want a top-level `login` command, not sure if it's here still from the pr (maybe cmdinit brings it in?)

michaeljguarino:
> @michaeljguarino I've moved CD commands to the top level. I had to move the plural login to the management group commands (duplicated with the CD login)

We should still keep the `cd` commands under the `cd` subcommand I think, that'll prevent the need to move the login command as well.

cmd/pluralctl/main.go
zreigz: yes, the `plural init` can handle login.

zreigz:
CD commands are in `cd` group again. The `plural init` can handle the login part, PTAL

PR #537, opened 2024-08-26
enhancement: remove /**/terraform/**/main.tf from default encryption filters

maciaszczykm:
It removes `/**/terraform/**/main.tf` and also `/**/terraform/**/main.tf*` from default enctryption filters.


michaeljguarino (commented):
Can we also write the gitattributes file only if it doesn't exist (and same w/ the gitignore file)?

I think it can get in the way of the git agent in the server if it causes a fs change

michaeljguarino (approved):

michaeljguarino (approved):

PR #566, opened 2024-10-07
better streamline plural up --cloud

zreigz:
## Summary
 - replace the cluster name field with querying all the console instances for a user, using a select to choose the one they want to use, and use that name for the cluster name

 - no longer ask for the az spread (just chose one as a default and ignore)

 - make the bucket prefix automatically plrl-cloud-${name}.

## Checklist

### DIFF
--- a/cmd/command/cd/cd_clusters.go
+++ b/cmd/command/cd/cd_clusters.go
@@ -118,19 +118,12 @@ func (p *Plural) cdClusterCommands() []cli.Command {
 }
 
 func (p *Plural) handleListClusters(_ *cli.Context) error {
-	if err := p.InitConsoleClient(consoleToken, consoleURL); err != nil {
-		return err
-	}
-
-	clusters, err := p.ConsoleClient.ListClusters()
+	clusters, err := p.ListClusters()
 	if err != nil {
 		return err
 	}
-	if clusters == nil {
-		return fmt.Errorf("returned objects list [ListClusters] is nil")
-	}
 	headers := []string{"Id", "Name", "Handle", "Version", "Provider"}
-	return utils.PrintTable(clusters.Clusters.Edges, headers, func(cl *gqlclient.ClusterEdgeFragment) ([]string, error) {
+	return utils.PrintTable(clusters, headers, func(cl *gqlclient.ClusterEdgeFragment) ([]string, error) {
 		provider := ""
 		if cl.Node.Provider != nil {
 			provider = cl.Node.Provider.Name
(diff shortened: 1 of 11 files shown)


cmd/command/up/up.go
```diff
@@ -119,3 +119,31 @@ func (p *Plural) handleUp(c *cli.Context) error {
 	utils.Highlight("Feel free to use terraform as you normally would, and leverage the gitops setup we've generated in the apps/ subfolder\n")
 	return nil
 }
+
+func getCluster(cd *cd.Plural) (id string, name string, err error) {
+	if cd == nil {
+		return "", "", fmt.Errorf("please provide a plural client")
```
michaeljguarino, line 125: probably should change this error message to something like:

```
You're cli is not logged into Plural, try running `plural login` to generate local credentials.
```

cmd/command/cd/cd_clusters.go
```diff
@@ -118,19 +118,12 @@ func (p *Plural) cdClusterCommands() []cli.Command {
 }
 
 func (p *Plural) handleListClusters(_ *cli.Context) error {
```
michaeljguarino, line 120: This actually does work, but we can also use the `consoleInstances` query, which just selects Plural cloud clusters.

pkg/provider/aws.go
```diff
@@ -159,17 +159,20 @@ func getEC2Client(ctx context.Context, region string) (*ec2.Client, error) {
 }
 
 // TODO: during Plural init we should ask the user to choose which AZs they want to use (first 3, random, manual, look at how CAPA does that). There should be a minimum limit of 3.
-func getAvailabilityZones(ctx context.Context, region string) ([]string, error) {
+func getAvailabilityZones(ctx context.Context, region string, cloudFlag bool) ([]string, error) {
```
michaeljguarino, line 162: could we just have this default to `first3` entirely?  It's not piped into `plural up` at all so is actually irrelevant at the moment.  Was only for CAPI if i remember

michaeljguarino (commented):
only a few small nits

cmd/command/up/up.go
zreigz: done

cmd/command/cd/cd_clusters.go
zreigz: I searched the whole `schema.graphql` and couldn't find this query

pkg/provider/aws.go
zreigz: done

michaeljguarino (approved):

PR #573, opened 2024-11-07
feat: add yaml overlay setting to plural pr template

floreks:
## Summary

Added YAML overlay feature to the `plural pr template`.

## Test Plan

Added unit tests

## Checklist


pkg/pr/updates.go
```diff
@@ -84,6 +91,70 @@ func processRegexReplacements(replacements []RegexReplacement, ctx map[string]in
...
+	if err := yaml.Unmarshal(base, &baseMap); err != nil {
+		return nil, err
+	}
+
+	if err := yaml.Unmarshal(overlay, &overlayMap); err != nil {
+		return nil, err
+	}
+
+	options := []func(*mergo.Config){mergo.WithOverride, mergo.WithSliceDeepCopy}
+	if merge == ListMergeAppend {
```
michaeljguarino, line 135: We should make this case insensitive (just `strings.ToUpper(merge)` it).  There's an annoying thing where internally in the elixir code this is downcased but it is presented via gql as uppercase that could cause bugs.

michaeljguarino: Oh wait i misread, it is being upcased

michaeljguarino (approved):

pkg/pr/updates.go
floreks: It should be already handled by https://github.com/pluralsh/plural-cli/pull/573/files#diff-93646044a7d487363a18f90d1c82940c54ca0961fdc2a290a4799884bfcc40eaR37-R51. When it will be converted from the CRD to this internal type, it will do case insensitive comparison.

PR #606, opened 2025-02-03
override default password

zreigz:
## Summary
Override `/oem/defaults.yaml` file to remove the kairos user's default password.

## Checklist

### DIFF
--- a/cmd/command/edge/image.go
+++ b/cmd/command/edge/image.go
@@ -113,9 +113,24 @@ func (p *Plural) handleEdgeImage(c *cli.Context) error {
 	}
 
 	utils.Highlight("building image\n")
+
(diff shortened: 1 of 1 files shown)


cmd/command/edge/image.go
```diff
@@ -113,9 +113,24 @@ func (p *Plural) handleEdgeImage(c *cli.Context) error {
 	}
 
 	utils.Highlight("building image\n")
+
+	// Override /oem/defaults.yaml file to get rid of the default password for the kairos user
+	dir, err := os.MkdirTemp("", "image")
```
maciaszczykm, line 118: We do not need to create this dir. You should use output dir instead.

PR #610, opened 2025-02-20
fix: Console URL parsing improvements

maciaszczykm:


michaeljguarino (approved):

michaeljguarino (commented):
we should probably get rid of that build dind image action to avoid the spammy failures

PR #611, opened 2025-02-25
custom helm chart url

kinjalh:
## Summary
For `plural cd install` and `plural cd cluster bootstrap` allow user to pass in path to helm chart for operator (optional parameter, if not passed in by user just use default).

Main purpose of this is so that users can bootstrap their clusters in airgapped environments

https://linear.app/pluralsh/issue/PROD-3235/allow-customizing-helm-chart-url-in-plural-cd-install-and-plural-cd

## Test Plan
- Ran a kubernetes cluster locally on my macbook
- Created a local `.tgz` file for the helm chart
- Passed it in as param when running `plural cd cluster bootstrap`
- Verified that my local cluster shows up on my console

### DIFF
--- a/pkg/console/agent.go
+++ b/pkg/console/agent.go
@@ -42,7 +42,7 @@ func IsAlreadyAgentInstalled(k8sClient *kubernetes.Clientset) (bool, error) {
 	return false, nil
 }
 
-func InstallAgent(url, token, namespace, version string, values map[string]interface{}) error {
+func InstallAgent(url, token, namespace, version, helmChartLoc string, values map[string]interface{}) error {
 	settings := cli.New()
 	vals := map[string]interface{}{
 		"secrets":    map[string]string{"deployToken": token},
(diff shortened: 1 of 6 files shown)


pkg/console/agent.go
```diff
@@ -42,27 +42,34 @@ func IsAlreadyAgentInstalled(k8sClient *kubernetes.Clientset) (bool, error) {
 	return false, nil
 }
 
-func InstallAgent(url, token, namespace, version string, values map[string]interface{}) error {
+func InstallAgent(url, token, namespace, version string, values map[string]interface{}, helmChartLoc string) error {
```
zreigz, line 45: you can add `helmChartLoc`  after `version` to have one group of string parameters

kinjalh: sounds good making revision

pkg/console/agent.go
```diff
@@ -42,27 +42,34 @@ func IsAlreadyAgentInstalled(k8sClient *kubernetes.Clientset) (bool, error) {
...
 
+	chartLoc := fmt.Sprintf("%s/%s", ReleaseName, ChartName)
+	if helmChartLoc == "" {
+		fmt.Println("Adding default Repo for deployment operator chart:", RepoUrl)
+		if err := helm.AddRepo(ReleaseName, RepoUrl); err != nil {
+			return err
+		}
+	} else {
+		fmt.Println("Using custom helm chart url:", chartLoc)
+		chartLoc = helmChartLoc
```
zreigz, line 66: The installation can fail if the chart is not added to the helm repo.  Did you test it with the fresh installation? This is the reason why we have  `helm.AddRepo` method

kinjalh: I did test using fresh installation. I thought we wanted to avoid trying to do `helm.AddRepo` when using a custom helm chart location since that won't work in airgapped environments?
correct me if i'm wrong though

zreigz: if it works, it's ok

zreigz (approved):

PR #616, opened 2025-03-10
fix: Various fixes and improvements

maciaszczykm:
Fixes PROD-3300 PROD-3293 PROD-3292.

Zrzut ekranu 2025-03-10 o 16 37 28


michaeljguarino (commented):
we can get rid of the `splitn` contract if it's not stable (not a huge deal)

michaeljguarino (commented):
i believe there was also a ticket to prevent non aws/gcp/azure providers from being used in `plural up` if we could knock that out

zreigz:
> i believe there was also a ticket to prevent non aws/gcp/azure providers from being used in `plural up` if we could knock that out

it's already done

pkg/manifest/manifest.go
```diff
@@ -113,20 +119,17 @@ func (pMan *ProjectManifest) ConfigureNetwork() error {
...
-			return err
-		}
 
 		if !strings.HasSuffix(res, pluralDomain) {
-			return fmt.Errorf("Not an onplural.sh domain")
+			res += "." + pluralDomain
+		}
+
+		if err := utils.ValidateDns(res); err != nil {
+			return err
```
maciaszczykm, line 132: Either `marcin` and `marcin.onplural.sh` will work.

pkg/provider/provider.go
```diff
@@ -69,13 +69,12 @@ func GetProvider() (Provider, error) {
 
 	provider := ""
 	prompt := &survey.Select{
-		Message: "Select one of the following providers:",
+		Message: "Select the cloud provider:",
```
michaeljguarino, line 72: I still don't see how this is restricting the set to just aws/gcp/azure though

PR #624, opened 2025-04-07
fix: plural edge download command

zreigz:
## Summary
 - Use the Go code to pull and unpack the edge image.
 - After the' docker login' command, the go code uses defaultKeychain to authenticate to the private repo.
 - Fix CVEs

## Checklist


michaeljguarino (approved):
linking against containerd has historically increased our CVE counts a bit, is there a way to do this w/o requiring it?  I still think we should be able to upload a standard tarball into the OCI registry.

zreigz:
> linking against containerd has historically increased our CVE counts a bit, is there a way to do this w/o requiring it? I still think we should be able to upload a standard tarball into the OCI registry.

should be possible. I will check it

PR #783, opened 2026-07-15
implement templated agentHelmValues

zreigz:
## Summary
 - Render agentHelmValues with Liquid when agentHelmValuesTemplateable is true, using cluster context (name, handle, id, metadata, tags, etc.).
 - Fetch deployment settings with backwards compatibility: try the full GraphQL query first, fall back to a minimal query (agentHelmValues + agentVsn).

## Checklist


pkg/console/agent_helm_values.go
```diff
@@ -0,0 +1,89 @@
...
+	settings *gqlclient.DeploymentSettingsFragment,
+	cluster *gqlclient.ClusterFragment,
+) (map[string]interface{}, error) {
+	if settings == nil || settings.AgentHelmValues == nil {
+		return map[string]interface{}{}, nil
+	}
+
+	raw := *settings.AgentHelmValues
+	rendered := raw
+	if lo.FromPtr(settings.AgentHelmValuesTemplateable) {
```
michaeljguarino, line 23: this field won't exist on a lot of console's yet, so if we fetch it, the gql query will error.  We need to ensure the code is robust to that error, and works backwards compatibly (maybe catch and requery w/o trying to select that field).

zreigz: It's already handled in this method:
```
func (c *consoleClient) GetGlobalSettings() (*gqlclient.DeploymentSettingsFragment, error) {
	resp, err := c.client.GetDeploymentSettings(c.ctx)
	if err != nil {
		return c.GetGlobalSettingsMinimal()
	}
	if resp == nil || resp.DeploymentSettings == nil {
		return c.GetGlobalSettingsMinimal()
	}
	return resp.DeploymentSettings, nil
}
```
It first tries to fetch the full DeploymentSettings. If that fails (or the response is nil), it falls back to `GetGlobalSettingsMinimal()`, which only fetches the two required fields. In that case, `AgentHelmValuesTemplateable` will be nil, and lo.FromPtr will return false.

michaeljguarino (approved):

